PR #1, opened 2025-09-05
Initial memtable-based implementation

tjgreen42:
Initial memtable-based implementation of Tapir PostgreSQL extension.

Implements full-text search with BM25 ranking using in-memory structures:
* Core access method with tpvector data type and <@> operator
* Shared memory string interning and posting list management
* Crash recovery via document ID page persistence
* Initial test suite with concurrency and memory limit validation
* Support for PostgreSQL text search configurations and BM25 parameters

This is the first deliverable in the design document that is part of this PR (v0.0a).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,183 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
svenklemm, line 7: branch should probably be renamed to main

sql/tapir--0.0.sql
```diff
@@ -0,0 +1,97 @@
...
+
+CREATE FUNCTION tp_handler(internal)
+RETURNS index_am_handler
+AS 'MODULE_PATHNAME', 'tp_handler'
+LANGUAGE C;
+
+CREATE ACCESS METHOD tapir TYPE INDEX HANDLER tp_handler;
+
+-- tpvector type
+
```
svenklemm, line 17: maybe run pgspot on this

sql/tapir--0.0.sql
```diff
@@ -0,0 +1,97 @@
...
+-- Access method
+
+CREATE FUNCTION tp_handler(internal)
+RETURNS index_am_handler
+AS 'MODULE_PATHNAME', 'tp_handler'
+LANGUAGE C;
+
+CREATE ACCESS METHOD tapir TYPE INDEX HANDLER tp_handler;
+
```
svenklemm, line 15: Have you considered pros/cons of dedicated schema instead of putting everything in public?

src/index.h
```diff
@@ -0,0 +1,145 @@
...
+/* Document ID page management for crash recovery */
+void		tp_add_docid_to_pages(Relation index, ItemPointer ctid);
+void		tp_recover_from_docid_pages(Relation index);
+
+/* Query limit tracking functions */
+void		tp_store_query_limit(Oid index_oid, int limit);
+int			tp_get_query_limit(Relation index_rel);
+void		tp_cleanup_query_limits(void);
+
+#endif							/* INDEX_H */
```
svenklemm:
```suggestion
```

src/index.h
```diff
@@ -0,0 +1,145 @@
...
+#include "commands/vacuum.h"
+#include "nodes/execnodes.h"
+#include "nodes/tidbitmap.h"
+#include "optimizer/cost.h"
+#include "optimizer/optimizer.h"
+#include "optimizer/pathnode.h"
+#include "storage/bufmgr.h"
+#include "storage/indexfsm.h"
+#include "utils/rel.h"
+#include "vector.h"
```
svenklemm: You probably don't need all of these in the header file and could move some in the c file but this is nitpicky.

scripts/format.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ ! -f "$TYPEDEFS_PATH" ]; then
+    echo "typedefs.list not found at $TYPEDEFS_PATH"
+    exit 1
+fi
+
+# Run pgindent
+$PGINDENT_PATH \
+  --indent="$INDENT_PATH" \
+  --typedefs="$TYPEDEFS_PATH" \
+  "$@"
```
svenklemm: Is this used?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,183 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        pg_version: [17]
```
svenklemm, line 14: what versions do you want to support with this? i saw some PG14 switches in c code

src/index.c
```diff
@@ -0,0 +1,2147 @@
...
+		return;
+	}
+
+	elog(DEBUG1,
+		 "Tapir cost estimation called for %d clauses, %d orderbys",
+		 list_length(path->indexclauses),
+		 list_length(path->indexorderbys));
+
+	/* Check for LIMIT clause and store it for optimization */
+	if (root && root->limit_tuples > 0 && root->limit_tuples < INT_MAX)
```
svenklemm, line 1360: Do you need to verify here that the LIMIT can actually be pushed down? I think this needs special testing to ensure this is handled correctly. You can only push down the limit if the scan produces output matching the order by of the query

src/index.c
```diff
@@ -0,0 +1,2147 @@
...
+	TimestampTz now;
+	TimestampTz cutoff;
+
+	if (!tp_query_limits_hash || !tp_shared_state)
+		return;
+
+	now = GetCurrentTimestamp();
+	cutoff = TimestampTzPlusSeconds(now, -TP_QUERY_LIMIT_TIMEOUT_SECONDS);
+
+	LWLockAcquire(tp_shared_state->query_limits_lock, LW_EXCLUSIVE);
```
svenklemm: Why does the query limit need to interact with the shared state? isnt this a backend local thing?

src/posting.c
```diff
@@ -0,0 +1,803 @@
...
+tp_get_or_create_posting_list(TpIndexState *index_state, const char *term)
+{
+	TpStringHashEntry *hash_entry = NULL;
+	TpPostingList *posting_list = NULL;
+	MemoryContext old_context;
+	size_t term_len;
+
+	if (!index_state->string_table)
+	{
+		elog(LOG, "tp_get_or_create_posting_list: index string_table is NULL");
```
svenklemm: should this be WARNING log level instead
https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS
if this is not an internal sanity check, you might wanna use ereport instead which allows you to also provide proper error code

src/posting.c
```diff
@@ -0,0 +1,803 @@
...
+
+	if (!term)
+	{
+		elog(LOG, "tp_get_posting_list: term is NULL");
+		return NULL;
+	}
+
+	term_len = strlen(term);
+
+	elog(LOG, "tp_get_posting_list: searching for term='%s'", term);
```
svenklemm: Thats gonna be superspammy, right? You probably want a different log level here

src/stringtable.c
```diff
@@ -0,0 +1,662 @@
...
+ * stringtable.c
+ *	  Custom hash table implementation for variable-length string interning
+ *
+ * This implements a simple hash table for string interning with
+ * table-level locking. Memory layout is designed for shared memory usage.
+ * Typically handles small strings (words/terms from text search processing).
+ *
+ * IDENTIFICATION
+ *	  src/stringtable.c
+ *
```
svenklemm: Have you considered using postgres hash table
https://github.com/postgres/postgres/blob/master/src/include/utils/hsearch.h

.github/workflows/ci.yml
tjgreen42: Right, plan is to support earlier versions than PG17.  (Exact ones TBD -- advice welcome).  I just turned this off in the CI matrix to simplify development for now.

src/posting.c
tjgreen42: Eww, you are correct.  Will fix.  (Eventually even most of the DEBUG log messages will go away, I just need them during the heavy development phase.)

src/stringtable.c
tjgreen42: Yes, that was what I used initially.  The problem is that it stores strings inline in entries, so you have to supply a maximum string length up front.  And terms are usually short, but don't have a maximum length.  So I didn't see a choice but to roll my own here.

src/posting.c
tjgreen42: Yes, good catch, and thanks for the pointer.  Could even be ERROR really, this should never happen.

src/index.c
tjgreen42: Yeah, ugh.  I really do need to know the LIMIT, but afaik this isn't directly exposed during evaluation, hence the ugly kludge.  Let me add some more tests probing correctness as you suggest.  I've also thought about how to bulletproof things even if we guess the LIMIT wrong, for instance, by reevaluating with a doubled LIMIT, then quadrupled, etc., but didn't implement any of that yet.

scripts/format.sh
tjgreen42: I have it running locally on my machine, but not in CI, since afaik you have to build all of Postgres to get pgindent.  I think it's probably better to nuke this thing and switch to clang-format like you guys have done in the main extension, but I didn't do that yet.

src/index.h
tjgreen42: Nits are okay and this is a fair point (probably in the other headers too).

sql/tapir--0.0.sql
tjgreen42: Hmm, hadn't considered.  I was just mimicking what I've seen other extensions like pgvector do.  I think we'd just be talking about putting the "internal" functions -- `tp_handler`, `tpvector_in` and friends -- in a subschema?  But many of them take `internal` as a parameter type anyway hence can't be called from SQL.

.github/workflows/ci.yml
tjgreen42: Ack

sql/tapir--0.0.sql
tjgreen42: Cool, TIL

.github/workflows/ci.yml
svenklemm: with timescaledb we are 3 major (15,16,17 atm). We drop oldest shortly after we add new one to keep the number the same, makes sense to do something similar here as platform follows similar pattern

sql/tapir--0.0.sql
tjgreen42: pgspot turns up some PS017 warnings (unqualified object reference), but these look innocuous since the schema search path will be locked down during extension creation.  And I double-checked that CREATE EXTENSION tapir SCHEMA foo seems to work properly.

src/index.h
tjgreen42: ?

src/index.c
tjgreen42: Yes, good catch.  Let me fix.

src/index.h
svenklemm: That was 2nd part of the `#pragma once` suggestion

tjgreen42: Got it.  Not sure what happened to the 1st part, I don't see it.   Agreed that `#pragma once` would probably be nicer -- I always forget that that's an option!

sql/tapir--0.0.sql
tjgreen42: Opting to leave this as is for now.

src/index.c
tjgreen42: I've added some smarter checks here along with test cases, but this is not entirely satisfactory.  Will revisit in future stages where the limit comes more into play.

src/stringtable.c
tjgreen42: Okay, I've done some fairly major revisions at this point to use DSM/DSA rather than traditional shared memory throughout the implementation, and I've changed the stringtable implementation to use `dshash` rather than a custom hash table implementation.  (`dshash` also requires fixed-sized keys, but I realized string pointers are perfectly fine fixed-sized keys).

PR #2, opened 2025-09-13
Query syntax revisions

tjgreen42:
Previously we were using `tpvector` for both queries and documents, with confusing/inconsistent parameter order to `<@>`.  This PR cleans things up by:
* Introducing a `tpquery` type for queries.  `tpvector` is now reserved for documents.  This aligns with `tsvector` / `tsquery` from regular Postgres FTS.
* The `<@>` variants now always take the document on the left (can be `text` or `tpvector`) and the query on the right (always `tpquery`)
* To allow more concise examples, the index name parameter to `tpquery` is optional.  It's not needed when index lookup is used; only when scoring is done row-at-a-time.

Examples of the new syntax:
```
-- Simple index scans: no index name required
SELECT *
FROM docs
ORDER BY content <@> 'search terms'::tpquery
LIMIT 10;

-- SELECT clause computes scores: now we need the index name
SELECT *, content <@> to_tpquery('search terms', 'docs_idx') AS score
FROM docs
ORDER BY content <@> 'search terms'::tpquery
LIMIT 10;
```

It is an `ERROR` to invoke `<@>` with a `tpquery` not containing an index name outside the context of an index scan.  Kind of a sharp edge admittedly since the planner's use of the index is not entirely predictable.  We may want to make the index name mandatory in the future for this reason -- it just makes examples look really cluttered!

There is a `WARNING` emitted the first time `<@>` is invoked in a query for scoring since it is quite a bit less efficient to score rows one at a time than to use the index.  (Unfortunately I don't believe the Postgres architecture lets us return scores used for `ORDER BY` to save this computation, if scores are really needed rather than just ranks.)

With these changes `tpvector` is arguably not really needed anymore (since unlike with Postgres FTS we index raw text fields rather than explicitly vectorized ones -- the vectorization is an internal index concern).  We can consider removing it in the future.  For now, it seems reasonable to leave it in.

Along with these changes to the syntax, the PR contains a variety of bugfixes and test improvements.

Note: performance has not been tuned at all.  The next PR will focus on performance, now that the syntax is more stable and basic correctness is reasonably well-established.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1142,10 +1142,6 @@ tp_execute_scoring_query(IndexScanDesc scan)
 	}
 	PG_CATCH();
 	{
-		ErrorData *errdata;
-
-		errdata = CopyErrorData();
-		FreeErrorData(errdata);
 		success = false;
 	}
 	PG_END_TRY();


src/index.c
```diff
@@ -1066,6 +1073,7 @@ tp_execute_scoring_query(IndexScanDesc scan)
 	}
 	PG_CATCH();
 	{
+		MemoryContextSwitchTo(oldcontext_func);
```
tjgreen42: @claude This looks like a no-op (switching to current memory context).  Can we just remove all of the context juggling here?

src/index.c
```diff
@@ -1103,15 +1142,10 @@ tp_execute_scoring_query(IndexScanDesc scan)
...
-		MemoryContext oldcontext = MemoryContextSwitchTo(so->scan_context);
-		ErrorData	 *errdata	 = CopyErrorData();
+		ErrorData *errdata;
 
-		elog(WARNING,
-			 "Exception during BM25 search for query '%s': %s",
-			 so->query_text,
-			 errdata->message);
+		errdata = CopyErrorData();
 		FreeErrorData(errdata);
```
svenklemm: you only need to copy if you want to rethrow the error

tjgreen42: Got it.  These try/catches are kind of questionable and will probably go away once the code here is fully stabilized.  (The scoring code is going to get more love once I start doing performance work today.)

PR #53, opened 2025-12-06
Syntax ergonomics and partitioned table support

tjgreen42:
## Summary

Adds ergonomic syntax for BM25 queries via implicit index resolution. Users can now write:

```sql
SELECT * FROM docs
ORDER BY content <@> 'search terms'
LIMIT 10;
```

Instead of the more verbose explicit form:

```sql
SELECT * FROM docs 
ORDER BY content <@> to_bm25query('search terms', 'docs_idx') 
LIMIT 10;
```

This also allows queries to work over partitioned tables (where each partition has its own named index).

## Implementation

- **Planner hook**: Transforms `text <@> text` to `text <@> bm25query` before planning, resolving the index OID from the column's BM25 index
- **Index OID storage**: Store index OID in bm25query instead of name string for efficiency  
- **OID cache**: Backend-local cache avoids repeated syscache lookups for extension OIDs (invalidated on extension install/drop)
- **Partitioned tables**: Parent index OID maps to partition indexes via `is_partition_of()` 

## Key behaviors

| Context | Behavior |
|---------|----------|
| ORDER BY | Index scan (fast, scores from index AM) |
| Standalone scoring | Explicit index lookup via planner hook |
| Multiple indexes on column | Warns user, picks first found |
| Partitioned tables | Partition indexes inherit from parent |

## Known limitations (documented in unsupported.sql)

1. **JOINs with implicit scoring** - Falls back to seq scan with standalone scoring. Returns all documents (non-matching get score 0) vs explicit `to_bm25query()` which uses the index and only returns matching documents.

2. **Multiple BM25 indexes on same column** - Planner hook picks the first matching index found. Use explicit `to_bm25query('query', 'index_name')` for control over which index is used.

3. **Expression indexes** - Not supported. Fails at index creation time with clear error message.

4. **Aggregates on scores** - Require explicit `to_bm25query()` syntax, which works correctly.

## Follow-up improvements (out of scope for this PR)

- **Double-scoring optimization**: Currently, queries that return scores in the SELECT list compute BM25 scores twice: once during index scan (for ordering) and once during projection (for the result column). This is a pre-existing issue and will be fixed in a future PR.

## Testing

- `implicit.sql` - Tests automatic index resolution
- `partitioned.sql` - Tests partitioned table support including sub-partitions  
- `unsupported.sql` - Documents known limitations with examples

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,16 +1,17 @@
-Source code in this repository is variously licensed under the Apache License
-Version 2.0, an Apache compatible license, or the Timescale License.
+The PostgreSQL License
 
-All source code should have information at the beginning of its respective file
-which specifies its licensing information.
+Permission to use, copy, modify, and distribute this software and its
+documentation for any purpose, without fee, and without a written agreement
+is hereby granted, provided that the above copyright notice and this paragraph
+and the following two paragraphs appear in all copies.
 
-* Outside of the "tsl" directory, source code in a given file is licensed
-  under the Apache License Version 2.0, unless otherwise noted (e.g., an
-  Apache-compatible license).
+IN NO EVENT SHALL TIGER DATA BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT,
+SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING LOST PROFITS, ARISING
+OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN IF TIGER DATA HAS
+BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-* Within the "tsl" folder, source code in a given file is licensed under the
-  Timescale License, unless otherwise noted.
-
-When built, separate shared object files are generated for the Apache-licensed
-source code and the Timescale-licensed source code. The shared object binaries
-that contain `-tsl` in their name are licensed under the Timescale License.
+TIGER DATA SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+THE SOFTWARE PROVIDED HEREUNDER IS ON AN "AS IS" BASIS, AND TIGER DATA HAS NO
+OBLIGATIONS TO PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR
+MODIFICATIONS.
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,15 @@
+pg_textsearch by Tiger Data (TM)
+
+Copyright (c) 2025-2026 Timescale, Inc. d/b/a Tiger Data. All Rights Reserved.
+
+Licensed under the PostgreSQL License (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   https://github.com/timescale/pg_textsearch/blob/main/LICENSE
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.


src/index.c
```diff
@@ -172,7 +173,7 @@ tp_auto_spill_if_needed(TpLocalIndexState *index_state, Relation index_rel)
 		MarkBufferDirty(metabuf);
 		UnlockReleaseBuffer(metabuf);
 
-		elog(DEBUG1,
+		elog(DEBUG2,
```
tjgreen42: Revert this back to DEBUG1

test/sql/basic.sql
```diff
@@ -46,14 +46,13 @@ INSERT INTO test_docs (content) VALUES
...
--- Test text <@> bm25query operator (should work)
+-- Test text <@> bm25query operator with explicit index for score projection
 SELECT content, content <@> to_bm25query('hello', 'test_tapir_idx') as score
-
 FROM test_docs
 ORDER BY score
 LIMIT 1;
 
--- Test index name mismatch error in index scan context
+-- Test non-existent index error (now caught at tpquery creation time)
```
tjgreen42: Don't leave historical markers like "now caught at"

test/expected/deletion.out
```diff
@@ -23,11 +23,11 @@ NOTICE:  BM25 index build started for relation deletion_idx
 NOTICE:  Using text search configuration: english
 NOTICE:  Using index options: k1=1.20, b=0.75
 NOTICE:  BM25 index build completed: 5 documents, avg_length=3.60, text_config='english' (k1=1.20, b=0.75)
--- Initial search - should find documents with "test"
+-- Initial search - should find documents with "test" (implicit index resolution)
```
tjgreen42: No need to say "implicit index resolution"

test/expected/inheritance.out
```diff
@@ -1,10 +1,12 @@
--- Test BM25 index behavior with table inheritance and partitioning
+-- Test BM25 index behavior with table inheritance
 -- Load pg_textsearch extension
 CREATE EXTENSION IF NOT EXISTS pg_textsearch;
 INFO:  pg_textsearch v0.0.6-dev: This is prerelease software and should not be used in production.
 INFO:  This release contains breaking changes in the bm25 index structure and will require existing indexes to be rebuilt.
+-- =============================================================================
 -- Test 1: PostgreSQL native table inheritance
--- Create parent and child tables
+-- BM25 indexes are not supported on tables with inheritance children
```
tjgreen42: I'm confused about this comment, does Test 1 demonstrate this?

test/expected/limits_1.out
```diff
@@ -0,0 +1,304 @@
...
+-- Test 5: LIMIT with OFFSET
+SELECT title, ROUND((content <@> 'performance')::numeric, 4) as score
+FROM limit_test
+ORDER BY 2
+LIMIT 5 OFFSET 2;
+         title         |  score  
+-----------------------+---------
+ Algorithm Design      | -1.8610
+ Information Retrieval |  0.0000
+ Text Processing       |  0.0000
```
tjgreen42: I can see from the zero scores that the index is not being used here, but I think these tests are trying to test LIMIT behavior *with* the index, not with standalone scoring?

test/expected/limits_1.out
```diff
@@ -0,0 +1,304 @@
...
+    SELECT title FROM limit_test
+    ORDER BY content <@> 'algorithm'
+    LIMIT 4
+) q3;
+NOTICE:  BM25 index scan: tid=(0,18), BM25_score=-1.3915
+ERROR:  no BM25 index found for text <@> text expression
+DETAIL:  Create a BM25 index on the column.
+HINT:  SELECT col <@> to_bm25query('q', 'idx') AS score
+-- Test 10: LIMIT pushdown safety verification
+-- These tests verify that LIMIT pushdown is only used when safe
```
tjgreen42, line 258: Same comment as earlier

test/expected/limits_1.out
```diff
@@ -0,0 +1,304 @@
...
+----------------------------------------------------------------------
+ Limit
+   ->  Sort
+         Sort Key: ((content <@> 'limit_test_idx:simple'::bm25query))
+         ->  Seq Scan on limit_test
+(4 rows)
+
+-- Case: Multiple ORDER BY clauses
+-- Tapir requires exactly one ORDER BY, so not using the index here is correct behavior
+-- Query plan varies by PG version (Sort vs Incremental Sort) - not testing EXPLAIN
```
tjgreen42, line 275: If there's a difference in behavior that is version-dependent, that is okay, we just have multiple expected outputs files

test/expected/merge.out
```diff
@@ -147,15 +147,15 @@ SELECT 'Phase 3: post-merge inserts (memtable + 1 L1 segment)' AS phase;
 
 -- Should find 3 documents with 'hello' (2 in L1, 1 in memtable)
 SELECT COUNT(*) AS hello_count_with_new_inserts FROM merge_test
-WHERE content <@> to_bm25query('hello', 'merge_test_idx') < 0;
+WHERE content <@> 'hello' < 0;
```
tjgreen42: We shouldn't have all these WHERE clauses in test cases if we expect the index to be used.  Can you check each use of WHERE and remove if the index appears expected to be used by the test?

src/query.h
```diff
@@ -4,25 +4,34 @@
...
+ * unresolved (InvalidOid) for later resolution by planner hooks.
  */
 typedef struct TpQuery
 {
 	int32 vl_len_;					   /* varlena header (must be first) */
-	int32 index_name_len;			   /* length of index name (0 if none) */
+	uint8 version;					   /* binary format version */
+	Oid	  index_oid;				   /* resolved index OID (InvalidOid if
+										* unresolved) */
 	int32 query_text_len;			   /* length of query text */
```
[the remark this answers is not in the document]
tjgreen42: Low priority: only 3 bytes padding, transient value not persisted to disk, FLEXIBLE_ARRAY_MEMBER already dominates struct size. Won't fix.

PR #78, opened 2025-12-17
Reclaim pages after segment compaction

tjgreen42:
## Summary
- Add page reclamation during segment compaction to prevent index bloat
- Pages from merged source segments are now freed to PostgreSQL's FSM
- New allocations check FSM first, reusing freed pages when available
- Demote spill/merge notices to DEBUG1 to reduce build noise

## Results (internal benchmark, 2.1M slack messages)

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Index Size | 831 MB | 473 MB | **43% smaller** |
| Pages Reused | 0 | 45,026 | ✓ |

### DIFF
--- a/test/expected/merge.out
+++ b/test/expected/merge.out
@@ -78,17 +78,13 @@ INSERT INTO merge_test (content) VALUES ('database world news');
 INSERT INTO merge_test (content) VALUES ('goodbye database friend');
 -- Second spill creates segment 2 in L0, which triggers merge to L1
 -- After merge: L0 is empty (cleared), L1 has merged segment
--- Enable DEBUG1 to verify merge messages
-SET client_min_messages = debug1;
+-- Note: Run with client_min_messages=debug1 to see merge DEBUG messages
 SELECT bm25_spill_index('merge_test_idx') IS NOT NULL AS second_spill;
-DEBUG:  Merging 2 segments at level 0
-DEBUG:  Merged 2 segments from L0 into L1 segment at block 6 (15 terms, freed 4 pages)
  second_spill 
 --------------
  t
 (1 row)
 
-RESET client_min_messages;
 -- After merge: merged segment in L1, L0 is empty
 SELECT 'Phase 2: after second spill + merge (L0 empty, 1 segment in L1)' AS phase;
                               phase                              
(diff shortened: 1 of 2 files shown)


src/segment/segment.c
```diff
@@ -1136,6 +1179,102 @@ tp_write_segment(TpLocalIndexState *state, Relation index)
...
+	BlockNumber		 page_index_block;
+	uint32			 i;
+
+	*pages_out = NULL;
+
+	reader = tp_segment_open(index, root_block);
+	if (!reader)
+		return 0;
+
+	/* Start with capacity for data pages + some extra for page index */
```
tjgreen42: How do we know the extra is sufficient?

src/segment/segment.c
```diff
@@ -1136,6 +1179,102 @@ tp_write_segment(TpLocalIndexState *state, Relation index)
...
+	capacity  = reader->num_pages + 16;
+	all_pages = palloc(sizeof(BlockNumber) * capacity);
+	num_pages = 0;
+
+	/* Collect all data pages from the page map */
+	for (i = 0; i < reader->num_pages; i++)
+	{
```
tjgreen42: PG style: elide curly braces here

src/segment/segment.c
```diff
@@ -1136,6 +1179,102 @@ tp_write_segment(TpLocalIndexState *state, Relation index)
...
+ */
+void
+tp_segment_free_pages(Relation index, BlockNumber *pages, uint32 num_pages)
+{
+	uint32 i;
+
+	for (i = 0; i < num_pages; i++)
+	{
+		/* Skip the metapage (block 0) - should never happen but be safe */
+		if (pages[i] == 0)
```
tjgreen42: Should be an ERROR

test/expected/merge.out
```diff
@@ -79,8 +79,6 @@ INSERT INTO merge_test (content) VALUES ('goodbye database friend');
 -- Second spill creates segment 2 in L0, which triggers merge to L1
 -- After merge: L0 is empty (cleared), L1 has merged segment
 SELECT bm25_spill_index('merge_test_idx') IS NOT NULL AS second_spill;
-NOTICE:  Merging 2 segments at level 0
```
tjgreen42, line 82: Should we turn on DEBUG logging just for this part of the test so that we can still confirm internal state?

PR #81, opened 2025-12-17
Implement V2 segment format with block storage for BMW optimization

tjgreen42:
## Summary
- Implements Phase 1 of block storage optimization with V2 segment format
- Block-based posting storage (128 docs/block) with skip index for future BMW optimization
- Fieldnorm table using Lucene SmallFloat encoding (1 byte per doc)
- Segment-local document IDs with CTID map for heap tuple lookup
- All segment writes switched to V2 format (breaking change, metapage version 5)

## Compatibility
- **Read**: Supports both V1 and V2 segments (query code branches on `header->version`)
- **Write**: V2 only - all new segments use V2 format
- **Upgrade path**: Existing indexes must be rebuilt (standard for pre-1.0 releases)

## Key Changes
- **New files**: `docmap.c/h` (CTID mapping), `fieldnorm.h` (length quantization)
- **segment_merge.c**: Complete rewrite to read V2 sources and write V2 output
- **segment_query.c**: V2 query iterator with block-based iteration
- **segment.c**: V2 writer and dump function updates

## Format Details
- `TpBlockPosting` (8 bytes): doc_id + frequency (down from 14-byte TpSegmentPosting)
- `TpSkipEntry` (16 bytes): per-block statistics for BMW block skipping
- `TpDictEntryV2` (12 bytes): skip_index_offset + block_count + doc_freq

## Configuration Changes
- **Memtable spill threshold**: Bumped from 800K to 32M posting entries (~1M docs/segment)
- **Benchmark config**: `segments_per_level = 16` for fewer merges during benchmarking
- **Segment statistics**: Added `bm25_summarize_index()` calls to benchmark load scripts

## Testing
- All 29 regression tests pass
- Merge, segment, and scoring tests validate V2 format correctness

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -179,6 +179,10 @@ jobs:
         export PATH="/usr/lib/postgresql/17/bin:$PATH"
         export PGDATABASE=bench_test
 
+        # Disable autovacuum to avoid profiling vacuum instead of queries
+        psql -c "ALTER SYSTEM SET autovacuum = off"
+        pg_ctl reload -D tmp_bench/data
+
         # Create a query script that runs many iterations for profiling
         cat > /tmp/profile_queries.sql << 'EOSQL'
         -- Run queries repeatedly to get meaningful profile samples
@@ -214,11 +218,11 @@ jobs:
         psql -f /tmp/profile_queries.sql &
         PSQL_PID=$!
 
-        sleep 1
-
-        # Find the postgres backend process
-        PG_PID=$(pgrep -f "postgres.*bench_test" | head -1)
-        echo "Profiling query PID: $PG_PID"
+        # Wait for backend to start, then find it
+        # With autovacuum disabled, we should only see our query backend
+        sleep 2
+        PG_PID=$(ps aux | grep "postgres:.*bench_test" | grep -v "autovacuum\|checkpointer\|walwriter\|bgwriter\|logical\|startup" | grep -v grep | awk '{print $2}' | head -1)
+        echo "Profiling query backend PID: $PG_PID"
 
         if [ -n "$PG_PID" ]; then
           # Record profile for query duration (up to 10 min)
@@ -246,6 +250,10 @@ jobs:
           echo "Could not find postgres backend PID for queries"
           wait $PSQL_PID || true
         fi
+
+        # Re-enable autovacuum
+        psql -c "ALTER SYSTEM RESET autovacuum"
+        pg_ctl reload -D tmp_bench/data
 
     - name: Generate profile summary
       run: |


src/segment/docmap.h
```diff
@@ -0,0 +1,98 @@
...
+	uint32 num_docs;   /* Number of documents assigned */
+	uint32 capacity;   /* Current capacity of arrays */
+
+	/* Output arrays (indexed by doc_id) */
+	ItemPointerData *ctid_map;	 /* doc_id → CTID */
+	uint8			*fieldnorms; /* doc_id → encoded length (1 byte) */
+} TpDocMapBuilder;
+
+/*
+ * Initialize a document map builder.
```
tjgreen42: Constructor

src/segment/docmap.h
```diff
@@ -0,0 +1,98 @@
...
+ * Call this before collecting documents.
+ */
+extern TpDocMapBuilder *tp_docmap_create(void);
+
+/*
+ * Add a document to the map.
+ * Returns the assigned doc_id (reuses existing ID if CTID already present).
```
tjgreen42, line 55: What happens if doc_length is different than previous?

src/segment/docmap.h
```diff
@@ -0,0 +1,98 @@
...
+ * doc_length is stored for fieldnorm encoding.
+ */
+extern uint32
+tp_docmap_add(TpDocMapBuilder *builder, ItemPointer ctid, uint32 doc_length);
+
+/*
+ * Look up doc_id for a CTID.
+ * Returns UINT32_MAX if not found.
+ */
+extern uint32 tp_docmap_lookup(TpDocMapBuilder *builder, ItemPointer ctid);
```
tjgreen42, line 66: Do we have a guard such that we never allow insertion of UINT32_MAX ctids, but only at most UINT32_MAX - 1?

src/segment/docmap.h
```diff
@@ -0,0 +1,98 @@
...
+
+/*
+ * Finalize the document map.
+ * Builds the ctid_map and fieldnorms arrays sorted by doc_id.
+ * After this call, the hash table is no longer needed.
```
tjgreen42, line 71: No longer needed as in caller should free?

src/segment/docmap.h
```diff
@@ -0,0 +1,98 @@
...
+ */
+extern void tp_docmap_finalize(TpDocMapBuilder *builder);
+
+/*
+ * Free the document map builder and all associated memory.
+ */
+extern void tp_docmap_destroy(TpDocMapBuilder *builder);
+
+/*
+ * Get the CTID for a doc_id (after finalize).
```
tjgreen42: If "after finalize" is actually important (not sure why it is), then we should have a flag in the struct and assert on this.

src/segment/segment_merge.c
```diff
@@ -285,22 +294,108 @@ merged_term_add_segment_ref(
 }
 
 /*
- * Initialize a posting merge source for streaming.
+ * Load a block of postings for V2 merge source.
```
tjgreen42: Say meaning of return values

src/segment/segment_merge.c
```diff
@@ -285,22 +294,108 @@ merged_term_add_segment_ref(
...
+				  (ps->current_block * sizeof(TpSkipEntry));
+	tp_segment_read(
+			ps->reader, skip_offset, &ps->skip_entry, sizeof(TpSkipEntry));
+
+	/* Ensure we have enough buffer space */
+	if (ps->skip_entry.doc_count > ps->block_capacity)
+	{
+		if (ps->block_postings)
+			pfree(ps->block_postings);
+		ps->block_postings = palloc(
```
tjgreen42, line 334: I don't understand, what happens to the previously-existing blocks?

PR #89, opened 2025-12-24
fix: buildempty() should write init fork

SteveLauC:
An index's `buildempty()` should modify the init fork, not the main fork. This commit replaces `ReadBuffer()` with `ReadBufferExtended(index, INIT_FORKNUM, P_NEW, RBM_NORMAL, NULL)` to fix the issue

Here is how to verify the bug using commit 20313076f2f43cab3f1a36be0f2805fea22bcf83:

```sh
$ cd pg_textsearch && git checkout 20313076f2f43cab
$ make install
$ psql
steve=# DROP EXTENSION IF EXISTS pg_textsearch;
steve=# create unlogged table test (content text);
steve=# create index idx_content on test using bm25 (content) with (text_config='english');
steve=# SELECT oid
FROM pg_class
WHERE relname = 'idx_content' AND relkind = 'i';
  oid  
-------
 36129
(1 row)

#-- exit psql

$ cd $PGDATA # cd to the PG data dir
$ find . -name '36129*'
./base/16384/36129_init
./base/16384/36129

$ ls -l ./base/16384/36129_init
.rw------- 0 steve 24 Dec 12:25 ./base/16384/36129_init
```

You can see that the init fork is empty, i.e., does not contain a meta page

### DIFF
--- a/test/expected/unlogged_index.out
+++ b/test/expected/unlogged_index.out
@@ -16,7 +16,7 @@ NOTICE:  Using index options: k1=1.20, b=0.75
 NOTICE:  BM25 index build completed: 1 documents, avg_length=2.00, text_config='english' (k1=1.20, b=0.75)
 -- VERIFICATION: Check if _init fork exists and has valid size
 --
--- We verify that the init fork size is exactly one block size (Metapage)
+-- We verify that the init fork contains at least the meta page
 SELECT 
     relname, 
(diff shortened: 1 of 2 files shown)


tjgreen42 (commented):
This looks like a nice catch.  What would a test case for this look like?

tjgreen42 (approved):
Cool, thanks for the unit test.  LGTM.

PR #91, opened 2025-12-24
chore: fix make test and override pgxs installcheck

SteveLauC:
This commit:

1. Adds the missing `test` target. It is mentioned in `make help` but has not been defined, when I run it locally, I see this:
   ```sh
   $ make test
   make: Nothing to be done for `test'.
   ```
   
2. Makes the `installcheck` target override the one defined by pgxs. Previously, we were not overriding it but appending new commands to it


SteveLauC:
Let me fix the CI

PR #102, opened 2026-01-01
Add Block-Max WAND (BMW) optimization for top-k queries

tjgreen42:
## Summary

- Implement Block-Max WAND (BMW) for top-k retrieval using block-level upper bounds in V2 segments
- Add top-k min-heap with O(1) threshold access and O(log k) updates for efficient result tracking
- BMW fast path for single-term queries (skip blocks that can't contribute to top-k)
- BMW fast path for multi-term queries with WAND-style doc-ID ordered traversal
- Batch doc_freq lookups to reduce segment open/close overhead for multi-term queries
- GUC variables `pg_textsearch.enable_bmw` and `log_bmw_stats` for debugging/benchmarking

## Performance (MS MARCO 8.8M docs, p50 latency)

| Query Length | pg_textsearch | System X | Result |
|--------------|---------------|----------|--------|
| 1 token | 10.14ms | 18.05ms | **1.8x faster** |
| 2 tokens | 12.54ms | 17.24ms | **1.4x faster** |
| 3 tokens | 15.14ms | 22.94ms | **1.5x faster** |
| 4 tokens | 19.72ms | 24.01ms | **1.2x faster** |
| 5 tokens | 25.79ms | 26.21ms | ~same |
| 6 tokens | 32.88ms | 33.47ms | ~same |
| 7 tokens | 42.09ms | 32.23ms | 1.3x slower |
| 8+ tokens | 63.41ms | 39.17ms | 1.6x slower |

**Cranfield:** 225 queries in 57ms (0.26 ms/query avg)

*Note: These benchmarks establish baselines for pg_textsearch—not a head-to-head comparison. System X has different defaults and tuning options; further iteration on configurations required.*

## Implementation Details

**New files:**
- `src/query/bmw.h` - Top-k heap, BMW stats, and scoring function interfaces
- `src/query/bmw.c` - Min-heap implementation, block max score computation, single-term and multi-term BMW scoring

**Key algorithm:**
1. Compute block max BM25 score from skip entry metadata (`block_max_tf`, `block_max_norm`)
2. Only score blocks where `block_max_score >= current_threshold`
3. Update threshold as better results are found
4. Memtable scored exhaustively (no skip index)

**Multi-term optimization:**
- Sort terms by IDF (highest first) for faster threshold convergence
- WAND-style doc-ID ordered traversal across terms' posting lists
- Batch doc_freq lookups: opens each segment once instead of once per term
- Reduces segment opens from O(terms × segments) to O(segments)

## Testing

- All regression tests pass
- Shell-based tests pass (concurrency, recovery, segment)
- Results match exhaustive scoring path with correct tie-breaking

### DIFF
--- a/test/expected/bmw.out
+++ b/test/expected/bmw.out
@@ -365,7 +365,6 @@ INSERT INTO bmw_hybrid (content) VALUES
 -- The memtable docs with higher tf should rank at top
 SELECT id, content <@> 'searchterm'::bm25query as score
 FROM bmw_hybrid
-WHERE content <@> 'searchterm'::bm25query < 0
 ORDER BY content <@> 'searchterm'::bm25query LIMIT 10;
   id  |         score          
 ------+------------------------
@@ -419,13 +418,13 @@ RESET pg_textsearch.index_memory_limit;
 -- Single-term
 WITH bmw AS (
     SELECT id, content <@> 'apple'::bm25query as score
-    FROM bmw_validate WHERE content <@> 'apple'::bm25query < 0
+    FROM bmw_validate
     ORDER BY content <@> 'apple'::bm25query LIMIT 20
 ),
 exhaustive AS (
     SELECT id, score FROM (
         SELECT id, content <@> 'apple'::bm25query as score
-        FROM bmw_validate WHERE content <@> 'apple'::bm25query < 0
+        FROM bmw_validate
         ORDER BY content <@> 'apple'::bm25query
     ) x LIMIT 20
 )
@@ -440,13 +439,13 @@ FROM (SELECT * FROM bmw EXCEPT SELECT * FROM exhaustive) diff;
 -- Multi-term
 WITH bmw AS (
     SELECT id, content <@> 'apple banana'::bm25query as score
-    FROM bmw_validate WHERE content <@> 'apple banana'::bm25query < 0
+    FROM bmw_validate
     ORDER BY content <@> 'apple banana'::bm25query LIMIT 20
 ),
 exhaustive AS (
     SELECT id, score FROM (
         SELECT id, content <@> 'apple banana'::bm25query as score
-        FROM bmw_validate WHERE content <@> 'apple banana'::bm25query < 0
+        FROM bmw_validate
         ORDER BY content <@> 'apple banana'::bm25query
     ) x LIMIT 20
 )
@@ -461,13 +460,13 @@ FROM (SELECT * FROM bmw EXCEPT SELECT * FROM exhaustive) diff;
 -- Three-term
 WITH bmw AS (
     SELECT id, content <@> 'apple banana cherry'::bm25query as score
-    FROM bmw_validate WHERE content <@> 'apple banana cherry'::bm25query < 0
+    FROM bmw_validate
     ORDER BY content <@> 'apple banana cherry'::bm25query LIMIT 20
 ),
 exhaustive AS (
     SELECT id, score FROM (
         SELECT id, content <@> 'apple banana cherry'::bm25query as score
-        FROM bmw_validate WHERE content <@> 'apple banana cherry'::bm25query < 0
+        FROM bmw_validate
         ORDER BY content <@> 'apple banana cherry'::bm25query
     ) x LIMIT 20
 )
@@ -509,7 +508,6 @@ WITH ranked AS (
            content <@> 'target'::bm25query as score,
            LAG(content <@> 'target'::bm25query) OVER (ORDER BY content <@> 'target'::bm25query) as prev_score
     FROM bmw_monotonic
-    WHERE content <@> 'target'::bm25query < 0
     ORDER BY content <@> 'target'::bm25query
     LIMIT 50
 )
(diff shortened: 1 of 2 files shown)


test/expected/bmw.out
```diff
@@ -0,0 +1,1168 @@
...
+ id |        score         
+----+----------------------
+  4 | -0.31600049138069153
+  1 |  -0.3106285035610199
+  2 | -0.30041441321372986
+  3 | -0.27344056963920593
+  7 | -0.24154625833034515
+(5 rows)
+
+-- Verify: id=4 (tf=4) should be first, then id=1 (tf=3), etc.
```
tjgreen42: You sure this is being checked in this test?

test/expected/bmw.out
```diff
@@ -0,0 +1,1168 @@
...
+ORDER BY content <@> 'exact'::bm25query LIMIT 10;
+ id |        score         
+----+----------------------
+  2 | -0.28768208622932434
+  3 | -0.28768208622932434
+  4 | -0.28768208622932434
+  5 | -0.28768208622932434
+(4 rows)
+
+-- Verify: IDs should be in ascending order (2,3,4,5) for tied scores
```
tjgreen42: Here again, not sure comment lines up with result

test/expected/bmw.out
```diff
@@ -0,0 +1,1168 @@
...
+    ('another memtable searchterm searchterm');
+-- Test: BMW should find results from both segment and memtable
+-- The memtable docs with higher tf should rank at top
+SELECT id, content <@> 'searchterm'::bm25query as score
+FROM bmw_hybrid
+WHERE content <@> 'searchterm'::bm25query < 0
+ORDER BY content <@> 'searchterm'::bm25query LIMIT 10;
+  id  |         score          
+------+------------------------
+ 1001 | -0.0007437646854668856
```
tjgreen42, line 371: Not sure these scores will be stable across HW / runs, consider rounding to fewer digits.  Also, why the `WHERE` clauses?  Ditch them please, here and elsewhere, unless you can defend their presence.

test/expected/bmw.out
```diff
@@ -0,0 +1,1168 @@
...
+    6 | -0.0004986796411685646
+    7 | -0.0004986796411685646
+    8 | -0.0004986796411685646
+(10 rows)
+
+-- Verify top results include the memtable docs (id > 1000)
+RESET pg_textsearch.index_memory_limit;
+DROP TABLE bmw_hybrid;
+-- ============================================================
+-- SECTION 10: Validate BMW equals exhaustive (LARGE SCALE)
```
tjgreen42, line 386: Why not use validation.sql?

tjgreen42: Same comment applies everywhere in this file -- use standard validation as we do elsewhere please.

test/expected/wand.out
```diff
@@ -0,0 +1,235 @@
+-- BMW Multi-term Scoring Bug Test
+--
+-- This test exposes a critical bug in multi-term BMW scoring.
```
tjgreen42: "critical bug" -- just remove this whole line

tjgreen42: Also, rephrase the whole thing.  If the bug is no longer present, since you fixed it, please rephrase and shorten all the verbiage.  Nobody cares about the prehistory.

tjgreen42 (commented):
Addressed review comments in e87f443:

**wand.out:3** - Rewrote header to be a regression test description, removed "critical bug" verbiage.

**bmw.out:40, 116** - Removed non-verified "Verify:" comments since they weren't being checked programmatically.

**bmw.out:374 (WHERE clauses)** - The WHERE clause `WHERE content <@> 'term'::bm25query < 0` is necessary: without it, non-matching documents would appear with score 0. The clause ensures we only get documents that actually match the query term.

**bmw.out:374 (score stability)** - Scores are deterministic based on BM25 formula with fixed k1/b parameters. The full precision is useful for detecting subtle scoring regressions.

**bmw.out:390 (validation.sql)** - The SECTION 10 approach compares BMW-optimized vs exhaustive paths using EXCEPT, which directly validates that both paths produce identical results. The validation.sql functions compare against a reference SQL implementation, which is a different (complementary) form of validation. Both are valuable; the current approach specifically tests the BMW optimization.

tjgreen42:
Correction on WHERE clauses:

After investigation, the WHERE clauses serve two different purposes:

1. **Non-existent query terms**: When the term doesn't exist in the index, the planner falls back to Seq Scan and returns all docs with score 0. WHERE is necessary to get empty results for these tests (e.g., SECTION 2 'elephant' test).

2. **Existing query terms**: Index Scan is used and only returns matching documents. WHERE is redundant but harmless.

So for correctness tests with matching queries, WHERE can be removed. For 'empty results' tests, WHERE must stay.

Given that removing WHERE from matching queries would be a large change with no functional benefit (and WHERE with Index Scan adds only a trivial Filter step), I'd suggest keeping them for consistency unless you feel strongly about removing them.

images/tapir_and_friends_v0.3.0-dev.png
SteveLauC, line 1: Just wanna say I love this Stranger Things-style banner 💯

tjgreen42: Aww, I was hoping someone would notice!

tjgreen42: BTW, the "not yet optimized" tagline is not entirely true anymore, as can be seen from the numbers in the PR description.  This little extension is starting to kick some butt.

PR #104, opened 2026-01-04
refactor: move buildphasename/spill_memtable to build.c and remove duplicates

SteveLauC:
Build phase constants `TP_PHASE_LOADING` and `TP_PHASE_WRITING` were defined in both `am/vacuum.c` and `am/build.c`, so this commit removes the definitions in `am/vacuum.c`.

It also moves `tp_buildphasename()` and `tp_spill_memtable()` from `am/vacuum.c` to `am/build.c` as they logically belong there.


src/am/vacuum.c
```diff
@@ -120,25 +116,6 @@ tp_vacuumcleanup(IndexVacuumInfo *info, IndexBulkDeleteResult *stats)
...
-		return "loading tuples";
-	case TP_PHASE_WRITING:
-		return "writing index";
-	default:
-		return NULL;
-	}
-}
-
 /*
  * tp_spill_memtable - Force memtable flush to disk segment
```
SteveLauC: BTW, I think we should move `tp_spill_memtable()` to `am/build.c` as well?

tjgreen42 (commented):
I accidentally pushed some commits directly to main, and your PR is an unfortunate victim as it now contains those commits as well.  (I've since backed out those commits).  Can you clean things up?  Thanks for the contributions!

src/am/vacuum.c
tjgreen42: Yes, that'd be a better home for it!

SteveLauC: Done

PR description and title updated to reflect this change

PR #114, opened 2026-01-07
Add parallel bulk index build support, part 1/2

tjgreen42:
## Summary

- Implement parallel CREATE INDEX using Postgres parallel infrastructure
- Each worker scans heap concurrently, builds local memtables, writes L0 segments
- Leader participates as worker 0; links all worker segment chains at end
- Compaction performed on leader node; part 2/2 will parallelize compaction

## Testing

- All 33 regression tests pass
- Falls back to serial when `max_parallel_maintenance_workers=0`

### DIFF
--- a/src/am/build.c
+++ b/src/am/build.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * am/build.c - BM25 index build, insert, and spill operations
--- a/src/am/build_parallel.c
+++ b/src/am/build_parallel.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * am/build_parallel.c - Parallel index build implementation
--- a/src/am/build_parallel.h
+++ b/src/am/build_parallel.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * am/build_parallel.h - Parallel index build structures
--- a/src/memtable/local_memtable.c
+++ b/src/memtable/local_memtable.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * memtable/local_memtable.c - Per-worker local memtable for parallel builds
--- a/src/memtable/local_memtable.h
+++ b/src/memtable/local_memtable.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * memtable/local_memtable.h - Per-worker local memtable for parallel builds
--- a/src/memtable/posting.h
+++ b/src/memtable/posting.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * posting.h - In-memory posting list structures for DSA shared memory
--- a/src/memtable/posting_entry.h
+++ b/src/memtable/posting_entry.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
  * Licensed under the PostgreSQL License. See LICENSE for details.
  *
  * memtable/posting_entry.h - Common posting entry type
--- a/src/mod.c
+++ b/src/mod.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Tiger Data, Inc.
+ * Copyright (c) 2025-2026 Tiger Data, Inc.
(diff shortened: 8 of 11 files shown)


tjgreen42:
Thanks for the review! I've addressed the feedback:

1. **term_len issue** (local_memtable.c:197): Fixed by introducing a separate `key_len` variable for the hash key lookup while preserving the original `term_len` for the posting.

2. **total_len accumulation** (local_memtable.c:276): Fixed by properly handling the case where an existing document length is updated, adjusting `total_len` by the difference between new and old values.

3. **NOTICE vs DEBUG1** (build.c:792): Keeping as NOTICE intentionally. Users benefit from knowing when parallel index build is active and how many workers are used. This matches the pattern in Postgres btree parallel builds. The DEBUG1 at line 801 is for a different case (explaining why parallel was *skipped*), which is less important for typical users to see.

.github/workflows/benchmark.yml
```diff
@@ -518,12 +518,9 @@ jobs:
   # in a separate PostgreSQL instance with identical configuration.
   system-x-benchmark:
     runs-on: ubuntu-latest
-    timeout-minutes: 180
-    # Run for msmarco or all datasets, and on scheduled runs
-    if: >
-      (github.event_name == 'schedule') ||
-      (github.event_name == 'workflow_dispatch' &&
-       (github.event.inputs.dataset == 'msmarco' || github.event.inputs.dataset == 'all'))
+    timeout-minutes: 360  # 6 hours for all datasets
```
tjgreen42: It's not 6 hours, check recent runs

src/memtable/local_memtable.c
```diff
@@ -0,0 +1,402 @@
...
+	entry = hash_search(memtable->term_hash, key, HASH_ENTER, &found);
+
+	if (!found)
+	{
+		/* Create new posting list */
+		oldcxt = MemoryContextSwitchTo(memtable->mcxt);
+
+		entry->posting			  = palloc(sizeof(TpLocalPosting));
+		entry->posting->term	  = pstrdup(term);
+		entry->posting->term_len  = strlen(term);
```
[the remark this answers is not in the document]
tjgreen42: The code already uses the passed `term_len` parameter correctly. Lines 209-215 copy exactly `term_len` bytes and store `term_len` in the posting structure. There's no strlen() recalculation - perhaps this was referring to an earlier version of the code.

src/memtable/local_memtable.c
```diff
@@ -0,0 +1,402 @@
...
+	entry = hash_search(memtable->doc_lengths, ctid, HASH_ENTER, &found);
+
+	if (!found)
+	{
+		entry->ctid = *ctid;
+		memtable->num_docs++;
+	}
+
+	entry->length = doc_length;
+	memtable->total_len += doc_length;
```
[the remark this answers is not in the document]
tjgreen42: This is already handled correctly in the current code (lines 293-297):
```c
else
{
    /* Adjust total length by the difference between new and old values */
    memtable->total_len += doc_length - entry->length;
}
```

src/am/build.c
```diff
@@ -738,6 +740,47 @@ tp_build(Relation heap, Relation index, IndexInfo *indexInfo)
...
+		 * the benefit. Note: reltuples may be -1 if table never analyzed.
+		 */
+#define TP_MIN_PARALLEL_TUPLES 100000
+
+		nworkers = plan_create_index_workers(
+				RelationGetRelid(heap), RelationGetRelid(index));
+
+		if (nworkers > 0 && reltuples >= TP_MIN_PARALLEL_TUPLES)
+		{
+			elog(NOTICE,
```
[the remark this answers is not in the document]
tjgreen42: Fixed in 7986bc0 - changed to DEBUG1.

.github/workflows/benchmark.yml
tjgreen42: Fixed in 7986bc0 - reduced to 60 minutes.

src/am/build.c
```diff
@@ -738,6 +740,108 @@ tp_build(Relation heap, Relation index, IndexInfo *indexInfo)
...
+		 * assumes ~160 bytes per row (reasonable for text search workloads).
+		 */
+#define TP_MIN_PARALLEL_TUPLES		100000
+#define TP_TUPLES_PER_PAGE_ESTIMATE 50
+
+		if (reltuples < 0)
+		{
+			BlockNumber nblocks = RelationGetNumberOfBlocks(heap);
+			reltuples = (double)nblocks * TP_TUPLES_PER_PAGE_ESTIMATE;
+			elog(DEBUG1,
```
tjgreen42: There's a lot of DEBUG trace in this PR, let's remove all such, unless there is a real need to keep.

src/am/build_parallel.h
```diff
@@ -0,0 +1,152 @@
...
+#define TP_PARALLEL_KEY_PAGE_POOLS	 UINT64CONST(0xB175DA7A00000003)
+#define TP_PARALLEL_KEY_TABLE_SCAN	 UINT64CONST(0xB175DA7A00000004)
+#define TP_PARALLEL_KEY_WAL_USAGE	 UINT64CONST(0xB175DA7A00000005)
+#define TP_PARALLEL_KEY_BUFFER_USAGE UINT64CONST(0xB175DA7A00000006)
+
+/*
+ * Per-worker segment tracking
+ *
+ * Each worker maintains its own chain of segments. After the build completes,
+ * the leader links all worker chains into L0.
```
tjgreen42: And then runs compaction?

src/am/build_parallel.c
```diff
@@ -0,0 +1,1368 @@
...
+	{
+		/* Link previous tail to new segment */
+		Buffer			 tail_buf;
+		Page			 tail_page;
+		TpSegmentHeader *tail_header;
+
+		tail_buf = ReadBuffer(index, my_info->segment_tail);
+		LockBuffer(tail_buf, BUFFER_LOCK_EXCLUSIVE);
+		tail_page				  = BufferGetPage(tail_buf);
+		tail_header				  = (TpSegmentHeader *)((char *)tail_page +
```
SteveLauC, line 155: I noticed that `TpSegmentHeader` starts from `page + SizeOfPageHeaderData` rather than `PageGetContents(page)` (`page + MAXALIGN(SizeOfPageHeaderData)`), is this done on purpose?

Practically, `SizeOfPageHeaderData` is usually 24 bytes, which is already MAXALIGN()'d, I am just curious

tjgreen42: Yeah, this is not consistent and should be cleaned up.  It only works because, as you note, 24 bytes is already a multiple of 8.  Lemme make a note.

PR #123, opened 2026-01-08
Fix bm25query varlena detoasting for binary I/O and scoring

tjgreen42:
## Summary

- Fix data corruption in bm25query type when Postgres uses short (1-byte) varlena headers
- Functions `tpquery_out`, `tpquery_send`, `bm25_text_bm25query_score`, and `tpquery_eq` now properly detoast varlena arguments
- Fix `tpquery_send` to use `pq_begintypsend()` for proper binary protocol formatting
- Fix `tpquery_recv` to use `pq_getmsgbyte()` for version field

## Background

When Postgres stores small varlena values, it may use a 1-byte header instead of the standard 4-byte header. Functions using `PG_GETARG_POINTER` instead of `PG_DETOAST_DATUM` would access struct fields at incorrect offsets, causing data corruption.

This manifested as:
- Truncated query text in `tpquery_out`
- Wrong index OID values
- Binary I/O failures with COPY BINARY

## Testing

- Added `binary_io.sql` test that verifies COPY BINARY round-trip for bm25query type

### DIFF
--- a/src/types/query.c
+++ b/src/types/query.c
@@ -223,7 +223,7 @@ tpquery_in(PG_FUNCTION_ARGS)
 Datum
 tpquery_out(PG_FUNCTION_ARGS)
 {
-	TpQuery	  *tpquery = (TpQuery *)PG_GETARG_POINTER(0);
+	TpQuery	  *tpquery = (TpQuery *)PG_DETOAST_DATUM(PG_GETARG_DATUM(0));
 	StringInfo str	   = makeStringInfo();
 
 	if (OidIsValid(tpquery->index_oid))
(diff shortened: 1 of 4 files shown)


src/types/query.c
```diff
@@ -276,8 +276,8 @@ tpquery_recv(PG_FUNCTION_ARGS)
 						 "an incompatible pg_textsearch version.",
 						 TPQUERY_VERSION)));
 
-	index_oid	   = pq_getmsgint(buf, sizeof(Oid));
-	query_text_len = pq_getmsgint(buf, sizeof(int32));
+	index_oid	   = pq_getmsgint(buf, 4);
```
tjgreen42: I still don't get why this is an improvement over the previous, same with next line

tjgreen42:
Addressed - reverted the `pq_getmsgint(buf, 4)` back to `pq_getmsgint(buf, sizeof(Oid))` and `sizeof(int32)`. You're right that the original was more self-documenting.

PR #154, opened 2026-01-17
feat: support expression indexes

### LINKED ISSUE #52: Support expression indexes
## Summary

Currently, BM25 indexes on expressions (e.g., `lower(content)`) are not supported and fail with a clear error message:

```sql
CREATE INDEX docs_lower_idx ON docs USING bm25(lower(content)) WITH (text_config='simple');
-- ERROR:  BM25 indexes on expressions are not supported
-- HINT:  Create the index on a column directly, e.g., CREATE INDEX ... USING bm25(content)
```

This ticket tracks the work needed to support expression indexes.

### 1. Index Build (`src/index.c`)

For simple column indexes, text is retrieved directly via `slot_getattr()` using the column number. For expression indexes, `ii_IndexAttrNumbers[0] = 0` (zero indicates an expression), requiring evaluation of `indexInfo->ii_Expressions` using `ExecInitExpr`/`ExecEvalExpr`.

### 2. Crash Recovery (`src/metapage.c`)

Recovery uses `heap_getattr()` to re-fetch text from heap tuples. For expressions, this doesn't work—we'd need to re-evaluate the expression against each recovered tuple.

### 3. Planner Hook Index Discovery (`src/planner_hook.c`)

The hook matches indexes by column number (`indkey.values[i] == attnum`). For expression indexes, indkey is 0. The hook would need expression tree comparison to match `lower(content) <@> 'query'` against the stored index expression.

### 4. Query Syntax

Users would need to write the exact expression in queries:
```sql
-- Column index
SELECT * FROM docs ORDER BY content <@> 'query';

-- Expression index (hypothetical)
SELECT * FROM docs ORDER BY lower(content) <@> 'query';
```

## Implementation Notes

Supporting expression indexes requires:
1. Expression evaluation context during index build
2. Expression evaluation during crash recovery
3. Expression tree matching in planner hook (comparing `Node*` trees via `equal()`)
4. Potentially storing expression metadata for comparison

## Workaround

For case-insensitive search, users can:
- Store pre-normalized text in a separate column
- Use `text_config='simple'` which doesn't apply stemming (though still case-sensitive tokenization)

SteveLauC:
Still WIP...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ PG_CPPFLAGS = -I$(srcdir)/src -g -O2 -Wall -Wextra -Wunused-function -Wunused-va
 # PG_CPPFLAGS += -DDEBUG_DUMP_INDEX
 
 # Test configuration
-REGRESS = aerodocs basic binary_io bmw compression coverage deletion vacuum dropped empty implicit index inheritance limits lock manyterms memory merge mixed parallel_build partitioned partitioned_many queries schema scoring1 scoring2 scoring3 scoring4 scoring5 scoring6 security segment segment_integrity strings unsupported updates vector unlogged_index wand text_config
+REGRESS = aerodocs basic binary_io bmw compression coverage deletion vacuum dropped empty expression_index implicit index inheritance limits lock manyterms memory merge mixed parallel_build partitioned partitioned_many queries schema scoring1 scoring2 scoring3 scoring4 scoring5 scoring6 security segment segment_integrity strings unsupported updates vector unlogged_index wand text_config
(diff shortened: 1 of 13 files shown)


SteveLauC:
Hi TJ @tjgreen42 , I’ve finished the design doc (see docs/expression_index.md). On the implementation side, everything is done except for the explicit index validation part, which needs us to merge #195 first.

It would be great if you take a look at it and give some feedback when you get a chance.

tjgreen42:
Closing in favor of #308

PR #163, opened 2026-01-22
refactor: simplify parallel build page allocation with 8-way merge

tjgreen42:
## Summary
- Simplifies parallel build page allocation to avoid smgr cache visibility issues
- Implements 8-way merge compaction enabling page reuse between merge batches
- Reduces index size regression from parallel builds

### 1. Simple pre-allocation strategy
Leader pre-allocates pages BEFORE workers start. Workers atomically claim pages from the pre-allocated range. This avoids smgr cache visibility issues where workers can't see pages extended by other workers.

Unused pages are truncated at the end of the build.

### 2. 8-way merge compaction
Merge only `tp_segments_per_level` (default 8) segments at a time instead of all segments. Pages freed from merged segments are recycled via FSM and available for subsequent merge batches. This reduces peak disk usage during large index builds.

### 3. Removed complexity
- Removed expansion factor GUC (`tp_parallel_build_expansion_factor`)
- Removed complex dynamic pool management
- Simplified page allocation logic in segment.c

## Testing
- All 39 regression tests pass
- All shell-based tests pass (concurrency, recovery, segment, stress)

### DIFF
--- a/src/am/build_parallel.c
+++ b/src/am/build_parallel.c
@@ -47,7 +47,6 @@ docmap_add_callback(ItemPointer ctid, int32 doc_length, void *arg)
 	tp_docmap_add(docmap, ctid, (uint32)doc_length);
 }
 
-
 /*
  * Worker build state - single memtable that gets spilled when full.
  */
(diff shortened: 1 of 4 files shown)


tjgreen42:
Note: The benchmark failure (`Page pool exhausted while writing page index`) is a **pre-existing issue on main** - the [most recent main benchmark](https://github.com/timescale/pg_textsearch/actions/runs/21237950822) also failed with the same error. This is unrelated to the FSM reclamation change in this PR.

The pool estimation for the Wikipedia 100K dataset needs to be addressed in a separate PR.

PR #167, opened 2026-01-22
Fix 'too many LWLocks taken' error when scanning hypertables

tjgreen42:
## Summary
Fixes "too many LWLocks taken" error when querying BM25 indexes on hypertables with many chunks.

### Problem
When scanning a hypertable with many chunks (e.g., 587), each chunk's BM25 index scan acquires a per-index LWLock but never releases it until transaction end, exceeding Postgres's ~200 lock limit.

### Solution
Release the lock immediately after `tp_memtable_search()` returns, since we've already extracted all CTIDs into memory.

This is the scan-path equivalent of the build-path fix in b7f6539.

## Testing
- Tested with 587-chunk hypertable (2.2M rows) - queries now succeed

### DIFF
--- a/src/am/scan.c
+++ b/src/am/scan.c
@@ -669,7 +669,7 @@ tp_execute_scoring_query(IndexScanDesc scan)
 	/* Find documents matching the query using posting lists */
 	success = tp_memtable_search(scan, index_state, query_vector, metap);
 
-	/* Release the lock - we've extracted all CTIDs we need */
+	/* Release the lock - we have extracted all CTIDs we need */
 	tp_release_index_lock(index_state);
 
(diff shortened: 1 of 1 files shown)


src/am/scan.c
```diff
@@ -669,6 +669,9 @@ tp_execute_scoring_query(IndexScanDesc scan)
 	/* Find documents matching the query using posting lists */
 	success = tp_memtable_search(scan, index_state, query_vector, metap);
 
+	/* Release the lock - we've extracted all CTIDs we need */
```
tjgreen42:
```suggestion
	/* Release the lock - we have extracted all CTIDs we need */
```

PR #181, opened 2026-01-27
Add upgrade paths from 0.4.x releases to 0.5.0-dev

tjgreen42:
## Summary
- Add upgrade SQL scripts: 0.4.0 → 0.4.1 → 0.4.2 → 0.5.0-dev
- Add bugfix upgrade path guidance to RELEASING.md
- Add reference to RELEASING.md in CLAUDE.md

## Testing
- `make test` passes

### DIFF
--- a/.github/workflows/upgrade-tests.yml
+++ b/.github/workflows/upgrade-tests.yml
@@ -147,15 +147,20 @@ jobs:
         psql -h /tmp -d upgrade_test -v expected_version="$EXPECTED_VERSION" << 'SQL'
         \set ON_ERROR_STOP on
 
+        -- Store expected version in temp table for use in DO blocks
+        CREATE TEMP TABLE _expected(version_prefix text);
+        INSERT INTO _expected VALUES (:'expected_version');
+
         -- Test 1: Verify extension upgraded to expected version
         DO $$
         DECLARE
           ver text;
-          expected text := :'expected_version';
+          expected text;
         BEGIN
+          SELECT version_prefix INTO expected FROM _expected;
           SELECT extversion INTO ver FROM pg_extension WHERE extname = 'pg_textsearch';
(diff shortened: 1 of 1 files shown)


RELEASING.md
SteveLauC, line 184: Should we include these bugfix versions in this matrix?

SteveLauC: Since we follow semantic versioning, there’s probably no need to include them—the version semantics already make this clear.

tjgreen42: We just need paths from any previous to current

.github/workflows/upgrade-tests.yml
SteveLauC: This `downloaded` output variable is defined but not used in the following steps. I think we can remove it as a download failure will cause the whole pipeline to exit (see the `exit 1` command in line 55)

SteveLauC: In line 65 and 280, the `cd ..` commands do not seem necessary. Every step has it own new shell that resets the process PWD

.github/workflows/upgrade-tests.yml
```diff
@@ -157,7 +154,7 @@ jobs:
           ver text;
         BEGIN
           SELECT extversion INTO ver FROM pg_extension WHERE extname = 'pg_textsearch';
-          IF ver NOT LIKE '0.4%' THEN
+          IF ver NOT LIKE '0.5%' THEN
```
SteveLauC: We are hard-coding the version info, which means we need to update this YAML file when we bump the version to the next dev version, which is actually fine.

But we can avoid this by extracting the version from the `pg_textsearch.control` file using the below command and use it in our SQL script (may need to trim the tailing new-line char):

```sh
$ awk -F"'" '/^[\t ]*default_version/ {print $2}' pg_textsearch.control
0.5.0-dev
```

PR #182, opened 2026-01-27
chore: drop directory prefix from the file identifiers

SteveLauC:
Drop directory prefix from the file identifiers so that we don't need to update them when moving files around.


tjgreen42 (changes requested):
Actually I'd prefer not to encode full paths, it's just another thing that needs maintaining when moving files around

PR #188, opened 2026-01-29
feat: parallel index build with leader-merges architecture

tjgreen42:
## Summary

- Implements parallel index build where workers scan heap and build memtables, leader merges and writes segments
- Workers use double-buffered memtables in shared DSA memory to avoid blocking
- Leader performs all writes using dynamic relation extension (P_NEW)
- Serial and parallel builds now produce identical index output (same size, same format)

## Architecture rationale

The key insight is moving all write responsibilities to the leader node:

1. **Sequential writes**: Leader writes segments sequentially, avoiding the heavy fragmentation that arose with earlier approaches where workers wrote their own segments
2. **CPU-bound workload**: Index building is mostly CPU-bound on the read path due to `to_tsvector` calls and memtable-population operations, so centralizing writes doesn't bottleneck throughput
3. **No followup compaction**: Previous approaches required a compaction step after workers finished; this design produces well-ordered segments directly

This architecture emerged from experimentation with several alternatives, informed by visualizations of page layout fragmentation.

## Key changes

- **Architecture**: Workers fill memtables → signal ready → leader merges N buffers into single segment
- **Threshold matching**: Uses same `tp_memtable_spill_threshold` (32M postings) as serial build
- **Compression**: Applies delta+bitpacking compression to posting blocks (same as serial)
- **Page boundary handling**: Fixed dictionary entry writing across page boundaries
- **Simplified page allocation**: Uses P_NEW instead of pre-allocated pool (no GUC needed)
- **Skip recovery docids**: Removed unnecessary docid writes during CREATE INDEX (atomic operation doesn't need recovery). This applies to both serial and parallel builds, so indexes are now identical.

## Benchmark (MS MARCO 8.8M docs)

| Configuration | Time | Index Size |
|---------------|------|------------|
| Serial | 4:27 | 1190 MB |
| Parallel (4 workers) | 1:26 | 1190 MB |

**3.1x faster** with proper configuration. Index sizes now match exactly.

Note: Postgres requires `maintenance_work_mem >= 64MB` for parallel builds (the planner won't assign workers otherwise).

## Testing

- `make installcheck` passes
- Tested on MS MARCO with 0, 1, 4 workers
- Index sizes match between serial and parallel builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,10 +234,14 @@ Postgres automatically uses parallel workers based on table size and configurati
 ```sql
 -- Configure parallel workers (optional, uses server defaults otherwise)
 SET max_parallel_maintenance_workers = 4;
+SET maintenance_work_mem = '256MB';  -- At least 64MB required for parallel builds
 
 -- Create index (parallel workers used automatically for large tables)
 CREATE INDEX docs_idx ON documents USING bm25(content) WITH (text_config='english');
 ```
+
+**Note:** The planner requires `maintenance_work_mem >= 64MB` to enable parallel index
+builds. With insufficient memory, builds fall back to serial mode silently.
 
 You'll see a notice when parallel build is used:
 ```


tjgreen42:
> Do we have a comparison between the old architecture (main branch) and this new one? From what I can see, the new architecture offers:
> 
> 1. Less random I/O as leader is the only writer, and it writes sequentially (mostly)
> 2. Less small tail segments
> 3. Less chances to trigger compaction

The main gist is to move write responsibilities to the leader node in order to make the writes mostly sequential (as you point out) and especially to avoid the heavy fragmentation that was arising with the previous approach, where workers wrote their own segments with a followup compaction step.  Since we are mostly CPU bound during indexing (and mostly on the read path) due to the calls to to_tsvector and memtable-population operations, this turns out to give pretty good speedups in the # of workers.  I had to experiment quite a bit to hit on this approach (several other PRs in flight that i'm now going to close) and add visualizations to see what was happening with the fragmentation.  Ironically, I intended the viz for human consumption, but the robot started using them immediately and became the main consumer.

src/am/build.c
```diff
@@ -786,8 +789,9 @@ tp_build(Relation heap, Relation index, IndexInfo *indexInfo)
 					 "Large table (%.0f tuples) with only %d parallel "
 					 "workers. "
 					 "Consider increasing max_parallel_maintenance_workers "
-					 "for "
-					 "faster index builds.",
+					 "and "
```
SteveLauC, line 790: Yeah, I think we should mention this in README.md as well, since this requirement is implicitly enforced by the planner and users may not be aware of it:

https://github.com/timescale/pg_textsearch/blob/80800439c2112008d6c4b573d5bb2de00e2efcb7/README.md?plain=1#L229-L236

.github/workflows/benchmark.yml
```diff
@@ -120,12 +120,14 @@ jobs:
...
         initdb -D tmp_bench/data --auth-local=trust --auth-host=trust
 
+        # Parallel build requires maintenance_work_mem >= 32MB * (workers + 1)
+        # 1GB allows up to 31 workers
+        NCPUS=$(nproc)
         cat >> tmp_bench/data/postgresql.conf << EOF
         unix_socket_directories = '$PWD/tmp_bench/socket'
         shared_buffers = 4GB
         maintenance_work_mem = 1GB
-        pg_textsearch.index_memory_limit = 512MB
```
SteveLauC, line 127: This legacy parameter still shows up frequently in the test scripts:

```sh
$ rg index_memory_limit
test/sql/memory.sql
11:SET pg_textsearch.index_memory_limit = 1;
63:RESET pg_textsearch.index_memory_limit;

test/expected/bmw.out
345:SET pg_textsearch.index_memory_limit = '64kB';
383:RESET pg_textsearch.index_memory_limit;
409:SET pg_textsearch.index_memory_limit = '64kB';
416:RESET pg_textsearch.index_memory_limit;
496:SET pg_textsearch.index_memory_limit = '64kB';
503:RESET pg_textsearch.index_memory_limit;
540:SET pg_textsearch.index_memory_limit = '64kB';
547:RESET pg_textsearch.index_memory_limit;
```

OPTIMIZATION_ROADMAP.md
```diff
@@ -18,16 +18,20 @@ Most high-impact query and indexing optimizations are now implemented:
 | Delta + bitpack compression | ✅ Done (v0.4.0) | 41% smaller indexes |
 | Fieldnorm quantization | ✅ Done (v0.2.0) | 1 byte vs 2-4 bytes per doc |
 | Doc ID mapping | ✅ Done (v0.2.0) | Compact u32 IDs, better compression |
-| Parallel index builds | 🔲 Not started | Faster bulk indexing |
+| Parallel index builds | ✅ Done (v0.5.0) | Faster bulk indexing |
 | SIMD decoding | 🔲 Not started | Minor speedup (~6% of query time) |
```
SteveLauC, line 22: Feel free to take this one, I’m not actively working on it. My current focus is getting more familiar with the codebase and finishing #154

src/am/build.c
SteveLauC: Maybe we should adjust the NOTICE message at line 811 as well

src/am/build_parallel.c
```diff
@@ -297,23 +135,19 @@ tp_build_parallel(
...
 	/* Ensure reasonable number of workers */
 	if (nworkers > TP_MAX_PARALLEL_WORKERS)
 		nworkers = TP_MAX_PARALLEL_WORKERS;
 
-	total_pool_pages = estimate_parallel_pool_pages(
-			RelationGetNumberOfBlocks(heap), nworkers);
-
 	/* Get snapshot for parallel scan */
 	snapshot = GetTransactionSnapshot();
 	snapshot = RegisterSnapshot(snapshot);
```
SteveLauC, line 145: We are registering snapshot regardless of the Postgres version. In serial build implementation, we only register/unregister snapshots for Postgres 18:

https://github.com/timescale/pg_textsearch/blob/80800439c2112008d6c4b573d5bb2de00e2efcb7/src/am/build.c#L527-L535

src/am/build_parallel.c
```diff
@@ -326,6 +160,11 @@ tp_build_parallel(
 
 	InitializeParallelDSM(pcxt);
 
+	/* Create DSA for shared memtable allocations */
+	dsa = dsa_create(LWTRANCHE_PARALLEL_HASH_JOIN);
```
SteveLauC: Using the `LWTRANCHE_PARALLEL_HASH_JOIN` tranche ID does not seem appropriate, maybe we should add a `TP_TRANCHE_PARALLEL_BUILD_DSA` or just re-use the existing `TP_TRANCHE_BUILD_DSA`?

src/am/build_parallel.c
```diff
@@ -416,952 +350,1758 @@ tp_init_parallel_shared(
...
+	/* Update corpus statistics in shared state */
+	pg_atomic_write_u64(&shared->total_docs, total_docs);
+	pg_atomic_write_u64(&shared->total_len, total_len);
+
+	/* Cleanup */
+	pfree(ready_buffers);
+}
+
+/*
+ * dshash parameters for string table (parallel-safe version)
```
SteveLauC: Looks like this is the doc comment for `tp_worker_create_string_table()`? And it uses tranche ID `TP_STRING_HASH_TRANCHE_ID` not `LWTRANCHE_PARALLEL_HASH_JOIN`

SteveLauC: BTW, in `constants.h`, we already defined tranche IDs for stringtable and doc length, which are not used:

https://github.com/timescale/pg_textsearch/blob/80800439c2112008d6c4b573d5bb2de00e2efcb7/src/constants.h#L77-L80

src/am/build_parallel.h
```diff
@@ -23,35 +31,66 @@
...
+	/* Memtable data (stored in shared DSA) */
+	dshash_table_handle string_hash_handle; /* Term -> posting list */
+	dshash_table_handle doc_lengths_handle; /* CTID -> doc length */
+
+	/* Statistics for this buffer */
+	int32 num_docs;		  /* Documents in this buffer */
+	int32 num_terms;	  /* Unique terms in this buffer */
+	int64 total_len;	  /* Sum of document lengths */
+	int64 total_postings; /* Total posting entries (for spill threshold) */
+	Size  memory_used;	  /* Approximate memory used */
```
SteveLauC: `num_terms` is unused, `build_dictionary_from_worker_buffer()` counts unique terms while iterating over the buffer string hash table.

`memory_used` is not used as well

src/am/build_parallel.c
```diff
@@ -416,952 +350,1758 @@ tp_init_parallel_shared(
...
+{
+	dshash_table	  *string_table;
+	dshash_seq_status  status;
+	TpStringHashEntry *entry;
+	TermInfo		  *terms;
+	uint32			   count	= 0;
+	uint32			   capacity = 1024;
+	dshash_parameters  params;
+
+	if (buffer->string_hash_handle == 0)
```
SteveLauC:
```suggestion
	if (buffer->string_hash_handle == DSHASH_HANDLE_INVALID)
```

The same pattern exists at lines 895, 1476, and 1515

tjgreen42:
> Did a brief review of the PR, sorry for the bunch of comments 🫠

Thanks much @SteveLauC for going through the PR, it's helpful.  I'd like to chat irl if you're game, ping me at tj@tigerdata.com.

Issue #194, opened 2026-02-02
to_bm25query with explicit index ignores index column


SteveLauC:
## Bug Description

When to_bm25query() is called with an explicit index name, the planner/executor uses that index regardless of whether it is built on the column referenced by the <@> operator.

```sql
-- This table has _two_ text columns
steve=# create table test_explicit_index (title text, content text);
CREATE TABLE

-- Index 'content_index' is built on column 'content'
steve=# create index content_index on test_explicit_index using bm25 (content) with (text_config='english');
CREATE INDEX

-- Search column `title` using an index built on a different column (`content`)
-- The index is still used.
steve=# explain select * from test_explicit_index order by title <@> to_bm25query('postgres', 'content_index');
                                  QUERY PLAN                                   
-------------------------------------------------------------------------------
 Sort  (cost=2261.84..2264.04 rows=880 width=72)
   Sort Key: ((title <@> 'content_index:postgres'::bm25query))
   ->  Seq Scan on test_explicit_index  (cost=0.00..2218.80 rows=880 width=72)
(3 rows)



steve=# insert into test_explicit_index values ('rust c java', 'postgres mysql');
INSERT 0 1

steve=# select * from test_explicit_index order by title <@> to_bm25query('postgres', 'content_index');
    title    |    content     
-------------+----------------
 rust c java | postgres mysql
(1 row)
```

## Environment

- **PostgreSQL version**: 17.7
- **pg_textsearch version**:  0.5.0-dev, the [latest commit](https://github.com/timescale/pg_textsearch/commit/60604649c2cef4b638c318a545bbbe6f72801d93)
- **Operating system**: macOS 26.0.1

## Steps to Reproduce

See the SQL snippets in the "Bug Description" section above
 
## Expected Behavior

The query should error out, since the <@> operator is applied to column
title but the explicitly specified index is built on column content.

## Actual Behavior

The explicitly specified index is used for the scoring and tokenization

## Additional Context

None

PR #195, opened 2026-02-02
fix: validate explicit index usage in to_bm25query()

### LINKED ISSUE #183: to_bm25query with explicit index name does not force planner to use that index
## Bug Description

When using to_bm25query('query', 'schema.index_name') with an explicit index name, the PostgreSQL planner may still
   choose a different BM25 index on the same column for the index scan. The specified index is only used for
  scoring/tokenization, not for physical access.

  This is problematic when you have multiple BM25 indexes on the same column with different text configurations
  (e.g., french vs simple), as the tokenization in the scanned index may not match the scoring index, leading to
  inconsistent or incorrect results.

## Environment

- **PostgreSQL version**: 17.7
- **pg_textsearch version**: 0.4.2
- **Operating system**: macOS 15.6

## Steps to Reproduce

```sql
  -- Setup

  DROP TABLE IF EXISTS documents;
  CREATE TABLE documents (
      id SERIAL PRIMARY KEY,
      category_id INT,
      content TEXT
  );

  -- Insert test data
  INSERT INTO documents (category_id, content)
  SELECT
      (i % 10),
      'Les ressources humaines gèrent les employés et leurs contrats de travail'
  FROM generate_series(1, 1000) i;

  -- Create two BM25 indexes with different configurations on the same column
  CREATE INDEX content_idx_french ON documents USING bm25(content) WITH (text_config = 'french');
  CREATE INDEX content_idx_simple ON documents USING bm25(content) WITH (text_config = 'simple');

  -- Analyze
  ANALYZE documents;

  -- Query explicitly requesting the French index
  EXPLAIN ANALYZE
  SELECT id, content <@> to_bm25query('ressources', 'public.content_idx_french') AS score
  FROM documents
  WHERE category_id = 1
  ORDER BY score
  LIMIT 10;
```
Which produced in my psql:
```
                                                                 QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.01..31.49 rows=10 width=12) (actual time=0.287..0.366 rows=10.00 loops=1)
   Buffers: shared hit=36
   ->  Index Scan using content_idx_simple on documents  (cost=0.01..314.80 rows=100 width=12) (actual time=0.287..0.365 rows=10.00 loops=1)
         Order By: (content <@> 'content_idx_french:ressources'::bm25query)
         Filter: (category_id = 1)
         Rows Removed by Filter: 81
         Index Searches: 0
         Buffers: shared hit=36
 Planning:
   Buffers: shared hit=37 read=6
 Planning Time: 1.853 ms
 Execution Time: 0.425 ms
```

## Expected Behavior

The query should use content_idx_french for the index scan since it was explicitly specified in to_bm25query.

## Actual Behavior

The planner may choose content_idx_simple for the index scan based on cost estimation, while still using
content_idx_french for BM25 scoring (as shown in the bm25query cast: 'content_idx_french:ressources'::bm25query).

Example EXPLAIN output showing the mismatch:

```sql
  Index Scan using content_idx_simple on documents
        Order By: (content <@> 'content_idx_french:ressources'::bm25query)
```

## Workaround

Adding a LIMIT inside a subquery can sometimes nudge the planner to use the correct index, but this is fragile and depends on cost estimates:

```sql
SELECT * FROM (
    SELECT id, content <@> to_bm25query('ressources', 'public.content_idx_french') AS score
    FROM documents
    WHERE category_id = 1
    ORDER BY score
    LIMIT 100  -- This changes planner behavior
) subq
WHERE score < -1;
```

### Suggested Enhancement

When to_bm25query is called with an explicit index name, the planner should prefer (or be forced to use) that index for the physical scan, not just for scoring calculations. This would ensure consistent tokenization between indexing and querying. A planner hint should be used ?

tjgreen42:
## Summary

Fixes #183 and #194.

When `to_bm25query('query', 'index_name')` specifies an explicit index, the planner now **uses that index** instead of picking arbitrarily from multiple BM25 indexes on the same column. This enables multi-lingual scenarios where different text configurations (e.g., french vs simple) are used on the same column.

### Issue #194: Wrong column validation
When an explicit index is specified but is on a different column than being searched, an ERROR is raised:
```
ERROR:  index "content_idx" is not on column "title"
DETAIL:  The explicitly specified index is not built on the column being searched.
```

### Issue #183: Explicit index forcing
When multiple BM25 indexes exist on the same column, the planner now respects the explicitly specified index:
```sql
-- Before: planner might choose idx_simple arbitrarily
-- After: planner uses idx_french as specified
SELECT * FROM docs 
ORDER BY content <@> to_bm25query('ressources', 'idx_french');
```

## Implementation

- **set_rel_pathlist_hook**: Swaps IndexPath's indexinfo when planner chose wrong BM25 index
- **Query tree walker**: Collects explicit index requirements, folding `to_bm25query()` FuncExpr to Const
- **Partitioned table support**: Walks up inheritance chain to allow child partition indexes

## Performance

Optimizations ensure minimal overhead for non-BM25 queries:

| Hook | Non-BM25 overhead |
|------|-------------------|
| post_parse_analyze | O(n) expression walk, fast per-node (type check + int compare) |
| planner | Skips plan tree walk if no BM25 operators found |
| set_rel_pathlist | Single NULL pointer check, returns immediately |

Key optimizations:
- `query_has_bm25_operators` flag tracks BM25 usage during parse analysis
- Planning context only set up when explicit index requirements exist
- Expensive syscache lookups in `plan_has_bm25_indexscan()` skipped for non-BM25 queries

## Testing

- New test file `test/sql/explicit_index.sql` covers both issues
- Updated `test/expected/mixed.out` for multi-index scenario

### DIFF
--- a/src/planner/hooks.c
+++ b/src/planner/hooks.c
@@ -438,49 +438,55 @@ transform_tpquery_opexpr(OpExpr *opexpr, ResolveIndexContext *context)
 	tpquery = (TpQuery *)DatumGetPointer(constNode->constvalue);
 
 	/*
+	 * The left operand of <@> must be a column reference for both index
+	 * resolution and index scan support.
+	 */
+	if (!IsA(left, Var))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+				 errmsg("left operand of <@> must be a table column"),
+				 errhint("Use column_name <@> bm25query, not a constant "
+						 "or expression.")));
+	}
+
+	/*
 	 * If the bm25query has an explicit index, validate that the index is on
 	 * the column being queried. This catches cases where the user specifies
 	 * an index that is on a different column.
 	 */
 	if (OidIsValid(tpquery->index_oid))
 	{
-		if (IsA(left, Var))
+		Var		  *var = (Var *)left;
+		Oid		   relid;
+		AttrNumber attnum;
+
+		if (get_var_relation_and_attnum(var, context->query, &relid, &attnum))
 		{
-			Var		  *var = (Var *)left;
-			Oid		   relid;
-			AttrNumber attnum;
-
-			if (get_var_relation_and_attnum(
-						var, context->query, &relid, &attnum))
+			if (!index_is_on_column(tpquery->index_oid, relid, attnum))
 			{
-				if (!index_is_on_column(tpquery->index_oid, relid, attnum))
-				{
-					char *index_name = get_rel_name(tpquery->index_oid);
-					char *col_name	 = get_attname(relid, attnum, false);
-					char *table_name = get_rel_name(relid);
+				char *index_name = get_rel_name(tpquery->index_oid);
+				char *col_name	 = get_attname(relid, attnum, false);
+				char *table_name = get_rel_name(relid);
 
-					ereport(ERROR,
-							(errcode(ERRCODE_WRONG_OBJECT_TYPE),
-							 errmsg("index \"%s\" is not on column \"%s\"",
-									index_name ? index_name : "(unknown)",
-									col_name ? col_name : "(unknown)"),
-							 errdetail(
-									 "The explicitly specified index is not "
-									 "built on the column being searched."),
-							 errhint("Use an index that is built on column "
-									 "\"%s\" of table \"%s\", or omit the "
-									 "index name to use automatic index "
-									 "resolution.",
-									 col_name ? col_name : "(unknown)",
-									 table_name ? table_name : "(unknown)")));
-				}
+				ereport(ERROR,
+						(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+						 errmsg("index \"%s\" is not on column \"%s\"",
+								index_name ? index_name : "(unknown)",
+								col_name ? col_name : "(unknown)"),
+						 errdetail(
+								 "The explicitly specified index is not "
+								 "built on the column being searched."),
+						 errhint("Use an index that is built on column "
+								 "\"%s\" of table \"%s\", or omit the "
+								 "index name to use automatic index "
+								 "resolution.",
+								 col_name ? col_name : "(unknown)",
+								 table_name ? table_name : "(unknown)")));
 			}
 		}
 		return NULL; /* Already resolved, validation passed */
 	}
-
-	if (!IsA(left, Var))
-		return NULL;
 
 	index_oid = find_index_for_var((Var *)left, context);
 	if (!OidIsValid(index_oid))
@@ -523,7 +529,16 @@ transform_text_text_opexpr(OpExpr *opexpr, ResolveIndexContext *context)
 	left  = linitial(opexpr->args);
 	right = lsecond(opexpr->args);
 
-	if (!IsA(left, Var) || !IsA(right, Const))
+	if (!IsA(left, Var))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+				 errmsg("left operand of <@> must be a table column"),
+				 errhint("Use column_name <@> 'query text', not a "
+						 "constant or expression.")));
+	}
+
+	if (!IsA(right, Const))
 		return NULL;
 
 	var		   = (Var *)left;
--- a/test/expected/explicit_index.out
+++ b/test/expected/explicit_index.out
@@ -1,7 +1,7 @@
 -- Test explicit index name in to_bm25query
 -- Validates fixes for GitHub issues #183 and #194
 CREATE EXTENSION IF NOT EXISTS pg_textsearch;
-INFO:  pg_textsearch v0.5.0-dev installed
+WARNING:  pg_textsearch v1.0.0-dev is a prerelease. Do not use in production.
 -----------------------------------------------------------------------
 -- Issue #194: to_bm25query with explicit index ignores index column
 -- Using an index on a different column should error
@@ -40,6 +40,10 @@ LIMIT 10;
   1 | -0.6931471824645996
 (1 row)
 
+-- This should ERROR: left operand is a constant, not a column
+SELECT 'hello' <@> to_bm25query('postgres', 'content_idx') AS score;
+ERROR:  left operand of <@> must be a table column
+HINT:  Use column_name <@> bm25query, not a constant or expression.
 -- Clean up
 DROP TABLE explicit_index_test;
 -----------------------------------------------------------------------
--- a/test/sql/explicit_index.sql
+++ b/test/sql/explicit_index.sql
@@ -34,6 +34,9 @@ FROM explicit_index_test
 ORDER BY score
 LIMIT 10;
 
+-- This should ERROR: left operand is a constant, not a column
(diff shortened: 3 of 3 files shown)


SteveLauC:
Thanks for the quick fix!

src/types/query.c
```diff
@@ -1101,6 +1126,20 @@ tpquery_has_index(TpQuery *tpquery)
 	return OidIsValid(tpquery->index_oid);
 }
 
+/*
+ * Check if tpquery has an explicitly specified index.
+ * Returns true only if user called to_bm25query(text, index_name).
```
SteveLauC: Except for `to_bm25query(text, index_name)`, A `TpQuery` with `TPQUERY_FLAG_EXPLICIT_INDEX` can also be created via `bm25query('index_name:text')`

src/planner/hooks.c
```diff
@@ -361,8 +436,47 @@ transform_tpquery_opexpr(OpExpr *opexpr, ResolveIndexContext *context)
 		return NULL;
 
 	tpquery = (TpQuery *)DatumGetPointer(constNode->constvalue);
+
+	/*
+	 * If the bm25query has an explicit index, validate that the index is on
+	 * the column being queried. This catches cases where the user specifies
+	 * an index that is on a different column (GitHub issue #194).
```
SteveLauC:
```suggestion
	 * an index that is on a different column.
```

Nit: Maybe we don’t need to reference the GitHub issue number in the code. Once the PR is merged, it’s not something we’ll care about

src/planner/hooks.c
```diff
@@ -800,7 +933,589 @@ replace_scores_in_plan(Plan *plan, BM25OidCache *oids)
...
+					 errhint("If this is not desired, use a planner hint to "
+							 "force a specific index, or use explicit "
+							 "to_bm25query('query', 'index_name').")));
+		}
+	}
+}
+
+/*
+ * Validate that when an explicit index is specified in bm25query, the planner
+ * uses that same index for the scan. This fixes GitHub issue #183.
```
SteveLauC:
```suggestion
 * uses that same index for the scan.
```

Same here

src/planner/hooks.c
```diff
@@ -73,13 +77,44 @@ static CacheState	oid_cache_state = CACHE_UNKNOWN;
 static BM25OidCache cached_oids;
 static bool			invalidation_registered = false;
 
+/*
+ * Structure to track explicit index requirements for a relation/column.
+ * Used during planning to force the planner to use the specified index.
+ */
+typedef struct ExplicitIndexRequirement
+{
+	Oid	 relid;				 /* Table OID */
```
SteveLauC: Current implementation only tracks table OID, we need to include an extra attribute number in order to work for queries 

```sql
select * from table
order by
    -- 2 indexes explicitly specified
    -- same table
    -- different columns 
    table.col1 <@> to_bm25query('query', 'index_col1'),
    table.col2 <@> to_bm25query('query', 'index_col2');
```

Even though we include an attribute number here, it won't work for this query:


```sql
select * from table
order by
    -- Same table, same column
    -- 2 indexes
    column <@> to_bm25query('query', 'column_index1'),
    column <@> to_bm25query('query', 'column_index2');
```

src/planner/hooks.c
```diff
@@ -73,13 +77,44 @@ static CacheState	oid_cache_state = CACHE_UNKNOWN;
...
+	Oid	 required_index_oid; /* The index that must be used */
+	bool is_explicit;		 /* true if from to_bm25query with index name */
```
SteveLauC: Why do we need this field? Currently, `collect_explicit_index_requirements()` collects all the resolved `TpQuery`s regardless of whether it has an explicitly specified index so we do need a flag to track the state. But looks like we can only collect explicit `TpQuery`s, then we can drop this field?

src/planner/hooks.c
```diff
@@ -800,7 +933,589 @@ replace_scores_in_plan(Plan *plan, BM25OidCache *oids)
...
+		return false;
+
+	if (IsA(node, OpExpr))
+	{
+		OpExpr *opexpr = (OpExpr *)node;
+
+		if (opexpr->opno == context->oid_cache->text_tpquery_operator_oid &&
+			list_length(opexpr->args) == 2)
+		{
+			Node *right = lsecond(opexpr->args);
```
SteveLauC, line 1141: We should probably check the left arg as well and ensure it is a `Var`

SteveLauC: Do you think it would make sense to have centralized query validation, for example in post_parse_analyze_hook? When we encounter an OpExpr(<@>), we could validate that:

1. the left arg is a Var node
2. the right arg is either a const `bm25query` node or a FuncExpr to_bm25query(...)

Then, when we walk the query tree later (e.g., in the planner), we can rely on these assumptions.

--------

The current implementation does not have such checks:

```sql
steve=# select * from test_explicit_index 
-- Left arg is a string constant
order by 'xx' <@> to_bm25query('rust', 'title_idx');
    title    |    content     
-------------+----------------
 rust c java | postgres mysql
(1 row)
```

src/types/query.c
tjgreen42: Fixed — updated the comment to mention both `to_bm25query(text, index_name)` and `bm25query('index_name:text')` as paths that set the explicit flag.

src/planner/hooks.c
tjgreen42: Done.

src/planner/hooks.c
tjgreen42: Done.

src/planner/hooks.c
tjgreen42: Added `attnum` to `ExplicitIndexRequirement` and changed `find_explicit_requirement_for_column()` to match on both relid and attnum. The `fix_bm25_indexpaths()` helper now gets the attnum from each IndexPath's indexinfo and finds the matching requirement per-column.

The same-table-same-column-two-indexes case remains unsupported, as you noted.

src/planner/hooks.c
tjgreen42: Agreed. Changed the walker to only collect explicit TpQuerys (`tpquery_is_explicit_index()` check moved inside the walker), which lets us drop the `is_explicit` field entirely.

src/planner/hooks.c
tjgreen42: Added — the walker now checks `IsA(left, Var)` and returns early (continues walking children) if the left arg isn't a column reference.

tjgreen42: Good idea. For this PR I've added the Var check in the collector (comment above), which prevents bogus planning requirements when the left arg isn't a column.

Centralized validation of `<@>` operand types (rejecting string constants on the left, etc.) would be a good follow-up. The `'xx' <@> to_bm25query(...)` case is technically valid at the operator level — it computes BM25 score of the constant text against the index — but it's almost certainly a user error. Worth a separate issue/PR to add a warning or error for non-Var left args in post_parse_analyze.

tjgreen42: Addressed in 6cbf1bf — both `text <@> bm25query` and `text <@> text` now error when the left operand is not a column reference.

PR #200, opened 2026-02-03
fix: implement CREATE INDEX CONCURRENTLY support

tjgreen42:
## Summary

- Implement proper `ambulkdelete` callback to report all indexed CTIDs during CIC validation
- Add SQL test verifying CIC works correctly without duplicate entries
- Add shell script tests for concurrent table operations during CIC

## Background

CREATE INDEX CONCURRENTLY requires the `ambulkdelete` callback to report all existing TIDs so `validate_index` can determine which rows need to be inserted. Without this, all rows would be re-inserted causing duplicates.

The fix adds two helper functions in `vacuum.c`:
- `tp_bulkdelete_memtable_ctids`: iterates through document lengths dshash to collect memtable CTIDs
- `tp_bulkdelete_segment_ctids`: iterates through segment document maps to collect segment CTIDs

## Testing

- SQL test `test/sql/concurrent_build.sql`:
  - Basic CIC index creation
  - Verification that index is marked valid
  - Verification that no duplicate entries are produced
  - DROP INDEX CONCURRENTLY
  - Regular CREATE INDEX for comparison

- Shell test `test/scripts/cic.sh`:
  - Concurrent INSERTs during CIC (100k documents)
  - Concurrent UPDATEs during CIC
  - Concurrent DELETEs during CIC
  - Mixed concurrent operations during CIC
  - Multiple concurrent CIC operations
  - Configured with lower spill thresholds to trigger multiple memtable spills

### DIFF
--- a/src/am/vacuum.c
+++ b/src/am/vacuum.c
@@ -175,17 +175,19 @@ tp_bulkdelete(
 
 		/* Iterate memtable CTIDs */
 		if (index_state != NULL)
-			tp_bulkdelete_memtable_ctids(index_state, callback, callback_state);
+			tp_bulkdelete_memtable_ctids(
+					index_state, callback, callback_state);
 
 		/* Iterate segment CTIDs at all levels */
 		for (int level = 0; level < TP_MAX_LEVELS; level++)
 		{
 			if (metap->level_heads[level] != InvalidBlockNumber)
 			{
-				tp_bulkdelete_segment_ctids(info->index,
-											metap->level_heads[level],
-											callback,
-											callback_state);
+				tp_bulkdelete_segment_ctids(
+						info->index,
+						metap->level_heads[level],
+						callback,
+						callback_state);
 			}
 		}
 	}


test/expected/coverage.out
```diff
@@ -1,7 +1,7 @@
 -- Additional coverage tests for untested code paths
 -- Tests debug functions, segment dump, and text<@>text operator
 CREATE EXTENSION IF NOT EXISTS pg_textsearch;
-NOTICE:  extension "pg_textsearch" already exists, skipping
```
SteveLauC, line 4: `test/sql/compression.sql` doesn’t `DROP EXTENSION pg_textsearch` in the cleanup section, so seeing an “extension already exists” message is expected. I’m curious why CI still passes after this change, though. 🫨

test/sql/concurrent_build.sql
```diff
@@ -0,0 +1,68 @@
...
+);
+
+INSERT INTO cic_basic (content) VALUES
+    ('the quick brown fox jumps over the lazy dog'),
+    ('postgresql is a powerful database system'),
+    ('full text search with bm25 ranking');
+
+-- Create index concurrently - should complete without error
+CREATE INDEX CONCURRENTLY cic_basic_idx ON cic_basic USING bm25(content)
+  WITH (text_config='english');
```
SteveLauC: Nit: I think it’s better to specify all configuration options when creating the index rather than relying on defaults. That way, if a default value changes, we won’t need to update the `expected/xxx.out` file

```
NOTICE:  Using index options: k1=1.20, b=0.75
```

src/am/vacuum.c
```diff
@@ -11,29 +11,148 @@
...
+	if (!doc_lengths_table)
+		return 0;
+
+	/* Iterate through all entries and call the callback for each CTID */
+	dshash_seq_init(&seq_status, doc_lengths_table, false);
+	while ((entry = dshash_seq_next(&seq_status)) != NULL)
+	{
+		if (ItemPointerIsValid(&entry->ctid))
+		{
+			/* Call the callback - returns true if tuple should be deleted */
```
SteveLauC:
```suggestion
			/* Call the callback */
```

This comment does not seem necessary here as we don't do vacuum in this function

PR #348, opened 2026-05-04
fix: chunked tokenization for oversized documents

tjgreen42:
## Summary
- Documents whose unique-token volume exceeds Postgres's `tsvector` 1 MB lexeme-dictionary cap (`MAXSTRPOS`) previously failed `CREATE INDEX`, `INSERT` (aminsert), `VACUUM` rebuild, `to_tpvector`, and standalone `<@>` scoring with `ERROR: string is too long for tsvector (N bytes, max 1048575 bytes)`.
- New `tp_tokenize_text` helper splits inputs >256 KB on ASCII whitespace (UTF-8-safe byte fallback when no whitespace is present), tokenizes each chunk via `to_tsvector_byid`, and merges per-chunk `(term, freq)` arrays via sort + collapse. Single-chunk fast path is unchanged.
- All five tokenization sites (`build.c` ×2, `build_parallel.c`, `vacuum.c`, `types/vector.c::to_tpvector`) route through the helper. The standalone scoring path in `bm25_text_bm25query_score` was also refactored to use the helper plus a sorted-array term lookup, since `find_term_frequency` previously walked the tsvector directly and would have re-hit the same 1 MB cap on seq-scan rows.
- Whitespace splitting is correct for whitespace-delimited scripts (Latin, Cyrillic, Greek, Arabic). Non-whitespace-delimited scripts (CJK, Thai) get byte/codepoint splits — acceptable because Postgres's default text-search parser doesn't emit per-word tokens for those scripts. Documented in README.

## Trigger
The bug requires *unique-token volume*, not raw byte size: `repeat('hello world ', 100000)` is 1.2 MB but only 2 lexemes and indexes fine. The repro and tests use ~250 K distinct tokens (~1.9 MB raw text).

## Testing
- New \`test/sql/large_documents.sql\` covers \`CREATE INDEX\`, \`INSERT\`, index-scan query, seq-scan / standalone-scoring query, and \`VACUUM\` on a 2 MB+ many-unique-token document.
- Full \`make installcheck\` (60 tests) passes locally on PG 17.


erimatnor:
@tjgreen42 For my education, how would you handle this with standard PostgreSQL tools (e.g., only `ts_vector`)? Would you have to split the document manually and create two `ts_vector`s? 

Is that what this function is doing automatically for you?

tjgreen42:
> @tjgreen42 For my education, how would you handle this with standard PostgreSQL tools (e.g., only `ts_vector`)? Would you have to split the document manually and create two `ts_vector`s?
> 
> Is that what this function is doing automatically for you?

Yep, you have to split the document manually, call `to_tsvector` on each chunk, and then combine the results.  And this is a bummer because it's hard to do the split in a language-neutral way.  We do support `TEXT[]` as an indexable type, though, so someone who knows how their documents should be split can do it for us in the source table.

Issue #342, opened 2026-04-29
bug: index queries fail on replicas


juris:
## Bug Description

We’ve encountered a bug where SELECT queries cause replicas to crash. I’m not a C expert, so I asked LLMs to review the pg_textsearch code and identify the root cause.
Here’s the bug report. Apologies if it’s a slop, but the issue is real: replicas are crashing with a segfault:

---
BM25 index queries fail on streaming replicas with `ERROR: invalid page index at block N`. Several segment-writing paths in `pg_textsearch` modify index pages via `MarkBufferDirty()` + `FlushRelationBuffers()` without `GenericXLog`. Primary persists pages via local buffer flush; replicas only apply WAL, so the pages never arrive. Reads on the replica hit stale FSM-recycled content and fail the magic check at `src/segment/segment.c:276-294`.

Affected paths:

| Path | Location | WAL-logged? |
|---|---|---|
| Serial `CREATE INDEX` / VACUUM rebuild — entire segment | `src/access/build_context.c:257` `tp_write_segment_from_build_ctx` | ❌ |
| Parallel `CREATE INDEX` leader — entire segment | `src/access/build_parallel.c:773` | ❌ |
| Memtable spill — page-index pages | `src/segment/segment.c:814` `write_page_index_internal` | ❌ |
| Level merge — page-index pages | `src/segment/segment.c:814` (called from `merge.c:1348`) | ❌ |
| Memtable spill / level merge — data pages | `segment.c:1472`, `merge.c:1654` | ✅ |
| Metapage updates | various | ✅ |

Page-index pages are allocated via `allocate_segment_page` (`segment.c:738`) directly, bypassing `writer.pages`. The post-hoc `GenericXLog FULL_IMAGE` pass in spill/merge iterates only `writer.pages`, so page-index blocks are skipped even where data pages are covered.

## Environment

- **PostgreSQL version**: 18.3 (streaming replication, CloudNativePG)
- **pg_textsearch version**: 1.1.0
- **Operating system**: CloudNativePG 1.29, Kubernetes 1.35, AWS Bottlerocket Linux 1.59.0

## Steps to Reproduce

1. Start primary, take base backup, start standby — standby streaming **before** the index is built.
2. On primary, build a BM25 index on a table large enough that the segment's page-index spans multiple blocks.
3. On standby (after replication catches up), run a top-k BM25 search using `<@> to_bm25query(...)` against that index.

Our deployment DDL on the primary:
```sql
[...]

PR #352, opened 2026-05-08
Pair WAL emission with buffer dirtying for segment pages

tjgreen42:
## Summary

Pair WAL emission with buffer dirtying for segment pages, so each page's
LSN is assigned before the buffer is unlocked and before any
flush/eviction path can observe it.

- New helper `tp_segment_log_dirty_buffer` (segment.c) — inside a
  critical section, marks the buffer dirty and (when
  `RelationNeedsWAL(index)`) emits `log_newpage_buffer(buf, false)`.
- Full-page-content writers (`tp_segment_writer_flush`,
  `write_page_index_internal`) call the helper at the same point they
  dirty the buffer.
- Small in-place backpatches (dict-entry `skip_index_offset` patches,
  segment-header finalization, `merge_sink_write_at`) emit
  GenericXLog **delta** records covering only the modified bytes,
  applied to pages that have already been WAL-logged in full by the
  writer flush above them.
- `allocate_segment_page` no longer does a speculative `PageInit` /
  `MarkBufferDirty`. The first content writer is responsible for
  `PageInit`, `pd_lower = BLCKSZ` (for content pages), marking dirty,
  and WAL emission.
- The previous post-hoc `GenericXLog FULL_IMAGE` sweeps in
  `build_context.c`, `build_parallel.c`, and `merge.c` are removed —
  replaced by per-page emission at content-write time.

## Motivation: correctness, not performance

`pg_textsearch` segment pages use custom page layouts and are updated
through direct buffer access. The previous code dirtied those buffers
first and emitted full-page WAL images later in a separate
end-of-segment sweep. That left a real window where a buffer was dirty
but its LSN did not yet cover its current contents. In vanilla
PostgreSQL this is hard to notice because the later sweep still makes
the final on-disk state recoverable, but the gap is observable:

1. **Flush or eviction before the deferred WAL pass.** Buffer eviction,
   checkpoint/writeback, or another storage layer can flush/evict a
   dirty page before the deferred sweep reaches it. The page is then on
   disk with no LSN, or with an LSN that does not cover the latest
   contents.
2. **Disaggregated or LSN-validating storage.** Storage systems that
   require every evictable page to have a valid LSN can reject or PANIC
   on such a page. This was observed downstream as a zero-LSN page
   eviction during ad-hoc BM25 index creation. The underlying issue is
   not specific to that environment — the page was dirty before
   WAL/LSN assignment.
3. **Crash/restart or replication timing around segment visibility.**
   The old code intended to make segment data durable before linking it
   from metadata, but the WAL record for a data page could be delayed
   relative to the first dirtying of that page. If a flush /
   checkpoint / storage interaction happened in that gap, recovery or
   replica state could depend on ordering the code did not explicitly
   enforce.

This patch closes the window by emitting WAL at the same point each
buffer is dirtied.

## Page-image flags

The full-page image uses `page_std = false`
(`log_newpage_buffer(buf, false)`):

- Writer-flush content pages set `pd_lower = BLCKSZ` (no hole), so
  REGBUF_STANDARD would be harmless but buys nothing.
- Page-index pages store entries between
  `pd_lower = SizeOfPageHeaderData` and `pd_upper = pd_special`.
  REGBUF_STANDARD would treat that range as a hole and zero it on
  replay, dropping the page-index contents. `page_std = false` logs
  the entire BLCKSZ unchanged.

## WAL-volume effect

This is a correctness fix; expect a small net **increase** in WAL
volume, not a decrease.

- `tp_segment_writer_flush` now emits one `log_newpage_buffer` record
  per page (≈ `BLCKSZ` + record header) instead of batching up to
  four page images per record in the deferred sweep. Net change:
  about +60 bytes of WAL record header per page.
- Dict-entry backpatches emit a GenericXLog delta record per
  dict-touched page in addition to the full-page image emitted earlier
  by `writer_flush`. The delta is small for sparse patches but can
  approach a full page when the dict-entry region is densely
  populated.
- Header finalization adds one small GenericXLog delta record.
- `merge_sink_write_at` emits one GenericXLog delta per chunk; on
  `main` it was `MarkBufferDirty` only, with the result folded into
  the deferred FPI sweep at no incremental WAL cost.

For a 64-page segment with ~1000 terms (so ~2 dict pages), the
increase is on the order of **~3 % (~16 KB) per segment** — negligible
relative to the segment data itself, and the price for getting WAL /
LSN ordering right.

There is no expected end-to-end performance improvement from this PR.
Query paths are unchanged; the build / merge paths do marginally more
WAL bookkeeping work.

## Tradeoffs

- WAL record count goes up. Per-record sizes are still dominated by
  the page images that have to be emitted anyway; record headers add
  a few tens of bytes per page.
- Caller discipline matters. The helper assumes callers hold the
  buffer exclusive lock and call it immediately after the final page
  modification, with no ERROR-capable work between the modification
  and the helper call. Backpatch sites use `GenericXLogStart` /
  `Finish`, which enforces the equivalent discipline.

## Why this approach

An alternative would be to rewrite the segment writers to use
GenericXLog page images as the primary mutation target throughout.
That is a larger refactor and still needs special care for custom page
layouts. This patch makes the existing buffer-mutation model safe by
ensuring each dirtying operation is paired with immediate WAL / LSN
assignment, while exploiting GenericXLog deltas where a delta is
genuinely smaller than a full page.

## Validation

- `git diff --check`
- `make` (clean, no warnings on changed files)
- `make installcheck` against PostgreSQL 18 — all 59 tests pass
- Focused code review of WAL / LSN ordering and custom page-layout
  handling

Not yet covered: the standby / replication suite from #342 / #343 has
not been re-run end-to-end against this branch. Worth wiring up before
merge.

### DIFF
--- a/src/access/build_context.c
+++ b/src/access/build_context.c
@@ -551,10 +551,16 @@ tp_write_segment_from_build_ctx(TpBuildContext *ctx, Relation index)
 	 * Write dictionary entries with real skip_index_offset
 	 * values. This must happen BEFORE tp_segment_writer_finish
 	 * so writer.pages is still valid.
+	 *
+	 * Logged as GenericXLog deltas (not full-page images) since
+	 * each batch only mutates a handful of 16-byte dict entries
+	 * in pages that the writer flush already WAL-logged in full.
 	 */
 	{
-		Buffer dict_buf		= InvalidBuffer;
-		uint32 current_page = UINT32_MAX;
+		Buffer			  dict_buf	   = InvalidBuffer;
+		Page			  dict_page	   = NULL;
+		GenericXLogState *xlog_state   = NULL;
+		uint32			  current_page = UINT32_MAX;
 
 		for (i = 0; i < num_terms; i++)
 		{
@@ -590,35 +596,35 @@ tp_write_segment_from_build_ctx(TpBuildContext *ctx, Relation index)
 			{
 				if (current_page != UINT32_MAX)
 				{
-					MarkBufferDirty(dict_buf);
+					GenericXLogFinish(xlog_state);
 					UnlockReleaseBuffer(dict_buf);
 				}
 				physical_block = writer.pages[logical_page];
 				dict_buf	   = ReadBuffer(index, physical_block);
 				LockBuffer(dict_buf, BUFFER_LOCK_EXCLUSIVE);
+				xlog_state = GenericXLogStart(index);
+				dict_page = GenericXLogRegisterBuffer(xlog_state, dict_buf, 0);
 				current_page = logical_page;
 			}
 
-			/* Write entry — handle page boundary spanning */
+			/* Write entry to working copy - handle page boundary spanning */
 			{
 				uint32 bytes_remaining = SEGMENT_DATA_PER_PAGE - page_offset;
 
 				if (bytes_remaining >= sizeof(TpDictEntry))
 				{
-					Page  page = BufferGetPage(dict_buf);
-					char *dest = (char *)page + SizeOfPageHeaderData +
+					char *dest = (char *)dict_page + SizeOfPageHeaderData +
 								 page_offset;
 					memcpy(dest, &entry, sizeof(TpDictEntry));
 				}
 				else
 				{
-					Page  page = BufferGetPage(dict_buf);
-					char *dest = (char *)page + SizeOfPageHeaderData +
+					char *dest = (char *)dict_page + SizeOfPageHeaderData +
 								 page_offset;
 					char *src = (char *)&entry;
 
 					memcpy(dest, src, bytes_remaining);
-					MarkBufferDirty(dict_buf);
+					GenericXLogFinish(xlog_state);
 					UnlockReleaseBuffer(dict_buf);
 
 					logical_page++;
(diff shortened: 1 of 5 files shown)


tjgreen42:
Thanks for the careful review! You correctly flagged the WAL-volume concern — newly allocated pages were getting an alloc-time full-page image *and* a writer-flush full-page image, and dict/header backpatches were emitting full-page images for pages that were already WAL-logged in full by the writer flush. I've now applied two mitigations directly to this PR:

**A) Drop the alloc-time full-page image.** `allocate_segment_page` no longer does `PageInit` / `pd_lower=BLCKSZ` / `MarkBufferDirty` / `tp_segment_log_dirty_buffer`. It just returns a block (FSM hit or `RBM_ZERO_AND_LOCK`), and the *first content writer* (`tp_segment_writer_flush` or `write_page_index_internal`) is responsible for `PageInit`, `pd_lower`, `MarkBufferDirty`, and WAL. Pages that are reached only by an error path that bails before flushing are left unmodified on disk — that's safe because nothing persisted references such pages.

**B) Use GenericXLog deltas for backpatches.** The dict-entry skip_index_offset patches in `tp_write_segment` / `tp_write_segment_from_build_ctx`, the segment-header finalization in both, and `merge_sink_write_at` now wrap their per-page mutations in `GenericXLogStart` / `GenericXLogRegisterBuffer(buf, 0)` / `GenericXLogFinish`. Each backpatch emits a *delta* covering only the modified bytes (typically a few `sizeof(TpDictEntry)`s or `sizeof(TpSegmentHeader)`), instead of another full-page image on top of the one the writer flush already produced.

The helper-doc comment in `segment.h` is updated to make the contract explicit: `tp_segment_log_dirty_buffer` is for full-content writers (writer flush, page-index write); small in-place updates should use GenericXLog deltas. The caller-discipline note remains — backpatch sites use the GenericXLog pair, which already enforces the no-error-between-mutate-and-finish invariant.

Net WAL volume should now be **lower** than `main` (pre-PR), not higher. PR description has a per-site breakdown table.

Verification: `make` clean (no warnings on changed files), `make installcheck` against PostgreSQL 18 — **all 59 tests pass**. The standby/replication suite from #342/#343 was not re-run end-to-end here; happy to wire that up in CI if it isn't already.

Force-pushed the amended commit (`2b533025` → `2475a887`).

PR #353, opened 2026-05-09
Restore standard installcheck behavior

tjgreen42:
## Summary
- remove the custom `installcheck` override so PGXS owns the target again
- keep SQL regression coverage available through `make installcheck` / `make test`
- keep shell-script coverage available through `make test-shell` and `make test-all`
- update Makefile help, CLAUDE.md, and test README to describe the target split

## Motivation
Downstream consumers expect PGXS `installcheck` to run SQL regression tests without triggering extension-specific shell scripts. The previous override made `installcheck` run both SQL regression and shell tests, which forced downstream harnesses to avoid the standard target.

Upstream shell-script coverage remains explicit through local targets and existing GitHub Actions steps.

## Validation
- `git diff --check`
- `make -n installcheck` shows PGXS pg_regress only
- `make -n test-all` shows SQL regression plus shell-script targets
- searched for stale docs claiming `installcheck` runs all tests
- code review + re-review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
         rm -rf tmp_check_shared
 
     - name: Run shell-based tests
-      timeout-minutes: 15
+      timeout-minutes: 25
       run: |
         export PATH="/usr/lib/postgresql/${{ matrix.pg_version }}/bin:$PATH"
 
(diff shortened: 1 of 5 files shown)


tjgreen42:
Good catch — the PR description's claim was indeed wrong with respect to `ci.yml`'s "Run shell-based tests" step. Fixed in e68199f4 by adding the missing scripts explicitly:

- **`ci.yml` main matrix "Run shell-based tests"** (PG17 + PG18): now invokes `recovery.sh`, `docid_chain_recovery.sh`, `shutdown_spill.sh`, `cic.sh`, `memory_accounting.sh`, and `multi_index.sh` alongside the existing scripts, using the same `✓`/`✗` pattern. Bumped `timeout-minutes` from 15 → 25 to absorb the extra work.
- **`ci.yml` sanitizer job** and **`sanitizer-build-and-test.yml`**: added the same six scripts after `recovery.sh`, mirroring the existing `timeout`-wrapped `bash -c` blocks. Restores the previous behavior where `make installcheck` implicitly ran all of `make test-shell` under sanitizer.
- **`ci.yml` PG17 coverage job**: added `multi_index.sh` alongside the other shell scripts; it was missing from the "all shell-based tests for coverage" list even before this PR.

`grep -rn 'multi_index' .github` now returns matches in both `ci.yml` (regular matrix, sanitizer job, and coverage job) and `sanitizer-build-and-test.yml`. PR description's "shell-script coverage remains explicit through ... existing GitHub Actions steps" is now accurate.

PR #357, opened 2026-05-11
fix: BMW infinite loop in block_max_skip_advance (#355)

### LINKED ISSUE #355: BMW scan hangs uninterruptibly on bucket-7 MS MARCO queries (Benchmarks workflow timing out for weeks)
## Symptom

The nightly **Benchmarks** workflow `insert-benchmark` job has been timing out at 6h consistently for weeks. Every failure hangs in the same step:

> **Step 19: Run MS MARCO concurrent insert queries**

The step runs `psql -f benchmarks/datasets/msmarco/queries.sql` after a 4-client pgbench concurrent insert workload of ~8.8M MS MARCO passages. Latency benchmarks for token buckets 1–6 each finish in <5 s; **bucket 7 (7-token queries) hangs forever**.

## Smoking-gun log line

From the May-8 run (`bdd3b317`):

```
2026-05-08T08:30:04Z   Token bucket 7 (7 search tokens):
2026-05-08T12:31:58Z   Received signal SIGTERM (runner shutdown)
```

→ **4 hours of total silence** from psql between the bucket-7 banner and the GH Actions runner being killed. No further log lines, no error.

## Affected runs

| Date | SHA | Run ID | Notes |
|---|---|---|---|
| 2026-04-27 | `35a2dc63` | [24980630180](https://github.com/timescale/pg_textsearch/actions/runs/24980630180) | pre-rmgr |
| 2026-05-01 | `142b32d8` | [25205596166](https://github.com/timescale/pg_textsearch/actions/runs/25205596166) | pre-rmgr |
| 2026-05-02 | `7f7bd41a` | [25245918436](https://github.com/timescale/pg_textsearch/actions/runs/25245918436) | pre-rmgr |
| 2026-05-03 | `7f7bd41a` | [25272307943](https://github.com/timescale/pg_textsearch/actions/runs/25272307943) | pre-rmgr |
| 2026-05-07 | `bdd3b317` | [25480813835](https://github.com/timescale/pg_textsearch/actions/runs/25480813835) | post-rmgr |
| 2026-05-08 | `bdd3b317` | [25540849097](https://github.com/timescale/pg_textsearch/actions/runs/25540849097) | post-rmgr |
| 2026-05-09 | `67af30f9` | [25594334436](https://github.com/timescale/pg_textsearch/actions/runs/25594334436) | post-rmgr |

So this is **not** a regression introduced by PR #347 (rmgr/replication) — it pre-dates that work — although the failure rate may have increased since (small samples).

`bdd3b317` ran successfully on 2026-05-06 and hung on 2026-05-07 + 08, so the bug is **intermittent / data-dependent** (same code, different segment topology produced by the random ordering of concurrent inserts).

## Why the hang is uninterruptible

`src/access/scan.c` and `src/scoring/bmw.c` contain **no `CHECK_FOR_INTERRUPTS` calls**:

```
$ grep -rn CHECK_FOR_INTERRUPTS src/access/scan.c src/scoring/
(no matches)
```

So once a backend enters the BMW top-k loop, neither `statement_timeout` nor `pg_cancel_backend()` nor `pg_terminate_backend()` can stop it. Only `kill -9` (SIGKILL) on the backend pid breaks the loop. This is itself a real bug worth fixing independently — even if we never reproduce the bucket-7 hang again, we do not want hung BM25 scans to be unkillable from SQL.

### Suggested cheap mitigation

Add `CHECK_FOR_INTERRUPTS()` at the top of the inner BMW loop in `src/scoring/bmw.c` and at every block-advance / posting-list advance in `src/access/scan.c`. Reference precedent: the merge code already does this (`src/segment/merge.c:1279`, `:1632`).

## Reproducing

Untested but the suspected recipe is:

```sh
# Build pg_textsearch with shared_preload_libraries set
# Configure: pg_textsearch.segments_per_level = 16   (matches benchmark.yml)
cd benchmarks/datasets/msmarco
./download.sh full      # 8.8M passages
psql -f insert/concurrent-setup.sql
ROWS=$(psql -tA -c "SELECT count(*) FROM msmarco_staging")
CLIENTS=$(nproc); TXN=$((ROWS / CLIENTS))
pgbench -f insert/pgbench-insert.sql -c $CLIENTS -j $CLIENTS -t $TXN -n
psql -f queries.sql     # → hangs around bucket 7 some fraction of the time
```

Likely needs several attempts due to data-dependent triggering.

## Next steps

1. ✅ #354 lands — next nightly hang produces a `diagnostics.log` artifact with `pg_stat_activity`, `pg_locks`, and **gdb stack traces** of the hung backend. That should pinpoint where in BMW the loop is stuck.
2. ⬜ Once we have a stack trace, root-cause the actual loop / lock / data-structure issue.
3. ⬜ Add `CHECK_FOR_INTERRUPTS()` to scan + BMW hot loops independently of (2).

## Related

- #354 — CI safety net so we get artifacts next time.
- #347 — rmgr/replication PR (NOT the cause but possibly correlated with frequency).
- #342 — earlier WAL-segment-write PR (NOT the cause but also pre-dates the bug).

tjgreen42:
## Summary

Fixes #355. The nightly **Benchmarks** workflow has been failing for weeks at the *Run MS MARCO concurrent insert queries* step. Last night's `diagnostics.log` showed pid 6339 spending 30+ minutes on bucket-8 with `wait_event=NULL` — a tight CPU spin, not a lock or I/O wait. This PR fixes the root cause plus two related observability gaps.

## Root cause

`score_segment_multi_term_bmw`'s WAND loop calls `block_max_skip_advance` when the block-max upper bound at the pivot can't beat the threshold. The old implementation picked `best_scorer` = pivot term with highest `max_score`, then seeked it to `min_block_end + 1` (the minimum block-end across pivot terms).

When `best_scorer` and the term that determined `min_block_end` are *different* terms, `best_scorer->cur_doc_id` can be greater than `seek_target`. In that case `seek_term_to_doc` is a no-op (the linear-scan-within-block loop sees `doc_id >= target` on the first iteration and returns without advancing). The outer WAND `while (active_count > 0)` loop then spins on the same state forever.

This becomes likely with high `term_count` (bucket 7/8 = 7-8 token queries) because `pivot_len > 1` and there are more chances for one pivot term to be in a later per-term block than another.

## Fix

Pick the pivot term whose current block ends *soonest* as the seek target. That term has `cur_doc_id ≤ min_block_end < min_block_end + 1 = seek_target` by construction, so the seek always advances. Defense-in-depth fallback to `advance_term_iterator` if the non-pivot cap collapses `seek_target` to the chosen term's current `doc_id` on a tie boundary.

The previous 'pick highest-max-score' selection was a performance heuristic; forward progress is the correctness requirement.

## CHECK_FOR_INTERRUPTS

Independently, the BMW + memtable hot loops in `src/scoring/bmw.c` contained zero `CHECK_FOR_INTERRUPTS`, so any future hang was uninterruptible from SQL — `statement_timeout` and `pg_cancel_backend()` had no effect, only SIGKILL. Added `CHECK_FOR_INTERRUPTS` to:

- The WAND main loop in `score_segment_multi_term_bmw`
- The block + posting loops in `score_segment_single_term_bmw`
- The segment iteration in `tp_score_{single,multi}_term_bmw`
- The memtable posting loops in `score_memtable_{single,multi}_term` (every 4096 postings, negligible overhead)

Precedent: `src/segment/merge.c` already calls `CHECK_FOR_INTERRUPTS` at `:1279` and `:1632`.

## benchmark.yml: install gdb

The *Capture diagnostics on failure* step gates gdb on `command -v gdb`, but gdb is not installed on `ubuntu-latest` runners by default — so previous failures silently produced no stack traces. Added `gdb` to the insert-benchmark install step.

## Verification

- ✅ Built clean on PG 18 (-O2 -g, no new warnings).
- ✅ All **59** regression tests pass via `pg_regress` against a freshly-initdb'd PG 18 instance with `shared_preload_libraries='pg_textsearch'`.
- ✅ `make format-check` clean with the CI-pinned `clang-format==21.1.8`.
- ⚠️ Deterministic SQL adversarial repro for the old buggy `block_max_skip_advance` was **not** constructed — BMW's pre-pivot pruning (`find_wand_pivot` returning false when no term combo beats threshold) keeps synthetic workloads from reaching the buggy code path. The production trigger is data-driven on ~8.8M MS MARCO passages with concurrent-insert segment topology. Internal HorizonDB tests at Microsoft were the original detection signal and will validate the fix.

Closes #355.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -82,7 +82,6 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y wget ca-certificates build-essential bc \
-            gdb \
             linux-tools-common linux-tools-generic linux-tools-$(uname -r) || true
         # Install FlameGraph for visualization
         git clone --depth 1 https://github.com/brendangregg/FlameGraph.git
@@ -1647,7 +1646,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y wget ca-certificates build-essential bc
+        sudo apt-get install -y wget ca-certificates build-essential bc gdb
         pip install wikiextractor
 
         # PostgreSQL
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -82,6 +82,7 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y wget ca-certificates build-essential bc \
+            gdb \
             linux-tools-common linux-tools-generic linux-tools-$(uname -r) || true
         # Install FlameGraph for visualization
         git clone --depth 1 https://github.com/brendangregg/FlameGraph.git
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ PG_CPPFLAGS += -Wno-unknown-warning-option -Wno-clobbered -Wno-packed-not-aligne
 # PG_CPPFLAGS += -DDEBUG_DUMP_INDEX
 
 # Test configuration
-REGRESS = abort aerodocs basic binary_io bmw bulk_load catalog_stats compression concurrent_build coverage deletion vacuum vacuum_bitmap vacuum_extended vacuum_rebuild dropped empty explicit_index expression_index force_merge implicit index inheritance limits lock manyterms max_shared_memory memory merge mixed parallel_build parallel_bmw partitioned partitioned_many partial_index pgstats queries quoted_identifiers rescan schema scoring1 scoring2 scoring3 scoring4 scoring5 scoring6 security segment segment_integrity strings temp_table text_array text_config unsupported updates vector vector_v1_rejected unlogged_index wand
+REGRESS = abort aerodocs basic binary_io bmw bmw_skip_advance bulk_load catalog_stats compression concurrent_build coverage deletion vacuum vacuum_bitmap vacuum_extended vacuum_rebuild dropped empty explicit_index expression_index force_merge implicit index inheritance limits lock manyterms max_shared_memory memory merge mixed parallel_build parallel_bmw partitioned partitioned_many partial_index pgstats queries quoted_identifiers rescan schema scoring1 scoring2 scoring3 scoring4 scoring5 scoring6 security segment segment_integrity strings temp_table text_array text_config unsupported updates vector vector_v1_rejected unlogged_index wand
 REGRESS_OPTS = --inputdir=test --outputdir=test
 
 PG_CONFIG = pg_config
--- a/src/scoring/bmw.c
+++ b/src/scoring/bmw.c
@@ -748,8 +748,9 @@ score_memtable_multi_term(
 			DocumentScoreEntry *entry;
 			bool				found;
 
-			/* Cheap CHECK every iteration is safe; this loop can be
-			 * very long when the memtable holds millions of postings. */
+			/* CHECK every 4096 postings; this loop can be very long
+			 * when the memtable holds millions of postings -- gating
+			 * amortizes the overhead. */
 			if ((i & 0xFFF) == 0)
 				CHECK_FOR_INTERRUPTS();
 
@@ -1295,17 +1296,21 @@ block_max_skip_advance(
 	{
 		if (!advance_term_iterator(terms[seek_term_idx]))
 			(*active_count)--;
+		if (stats)
+			stats->blocks_skipped++;
 	}
-	else if (!seek_term_to_doc(terms[seek_term_idx], seek_target))
-		(*active_count)--;
+	else
+	{
+		if (!seek_term_to_doc(terms[seek_term_idx], seek_target))
+			(*active_count)--;
+		if (stats)
+		{
+			stats->blocks_skipped++;
+			stats->seeks_performed++;
+		}
+	}
 
 	restore_ordering(terms, term_count, seek_term_idx);
-
-	if (stats)
-	{
-		stats->blocks_skipped++;
-		stats->seeks_performed++;
-	}
 }
 
 /*
--- a/src/scoring/bmw.c
+++ b/src/scoring/bmw.c
@@ -533,8 +533,6 @@ score_segment_single_term_bmw(
 		{
 			float4 score;
 
-			CHECK_FOR_INTERRUPTS();
-
 			/*
 			 * Break if iterator auto-advanced to next block.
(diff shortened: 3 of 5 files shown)


tjgreen42:
Thanks for the careful review @claude — all three flagged correctly. Addressed in 0dbe26b3:

**🔴 gdb in wrong job** — You're right, I added it to the `profile` job (`workflow_dispatch` only) instead of `insert-benchmark` (the one that actually runs on `schedule` and has the diagnostics step). Moved `gdb` from the profile install step to the insert-benchmark install step at line 1649. The other jobs (paradedb, full, baseline, validate-compression) don't have `Capture diagnostics on failure` steps so I left them alone.

**🟡 seeks_performed over-counts** — Moved the increment inside the seek branch only. `blocks_skipped` still counts both branches (it's the right grain for that metric — we did skip a block either way), but `seeks_performed` now matches its documented contract ("Binary search seeks executed").

**🟡 4096-posting CHECK comment** — Reworded to "CHECK every 4096 postings; this loop can be very long when the memtable holds millions of postings — gating amortizes the overhead."

Also added `test/sql/bmw_skip_advance.sql` (coverage smoke test that exercises `block_max_skip_advance` 4856 times against the patched code, bounded by 30s `statement_timeout` as the regression backstop). Full 59-test regression suite + `format-check` clean against PG 18 locally.

tjgreen42:
@claude can you take another pass — want to make sure my response to your previous review didn't introduce new issues

tjgreen42:
More follow-ups I caught on a self-review pass:

- **Redundant inner CHECK_FOR_INTERRUPTS removed.** `score_segment_single_term_bmw` had a CHECK_FOR_INTERRUPTS in the inner per-posting while-loop, but the outer per-block for-loop already calls it. The inner loop is bounded by `TP_BLOCK_SIZE = 128` postings, so the outer check is more than dense enough. Dropped the inner one to avoid the ~128× CHECK overhead on hot single-term scans.
- **Memtable BMW path now covered by the test.** Previous version of `bmw_skip_advance.sql` only exercised the segment BMW paths (everything went into segments via CREATE INDEX). The CHECK_FOR_INTERRUPTS calls I added to `score_memtable_single_term` and `score_memtable_multi_term` were among the uncovered lines codecov was complaining about. Extended the test to insert fresh docs after CREATE INDEX without spilling, so the next queries traverse both segments *and* the memtable.

The auto-reviewer was out of API credits when I asked for another pass, but these are the issues I'd expect it to flag.

PR #367, opened 2026-05-12
fix(bmw): include non-pivot term max scores in block-skip safety check (#365)

### LINKED ISSUE #365: BMW reports under-scored results for some docs when K (default_limit) is much larger than the actual LIMIT
## Summary

`tp_score_multi_term_bmw` (and the per-segment `score_segment_multi_term_bmw`) computes systematically too-low scores for some docs when called with a large K (e.g., `pg_textsearch.default_limit` = 1000) compared to the actual application LIMIT (e.g., 10). Same doc, same data, same backend session — only K differs.

## Reproducer

Local pg17 + concurrent-pgbench-built 8.8M-passage MS MARCO corpus.

### Query A: LIMIT 10 propagates to BMW (K=10)

```sql
SET pg_textsearch.log_scores = on;
SELECT passage_id, -(passage_text <@> to_bm25query('3 ways a log can move when bucking on side hill', 'msmarco_bm25_idx'))::float8 as score
FROM msmarco_passages
ORDER BY passage_text <@> to_bm25query('3 ways a log can move when bucking on side hill', 'msmarco_bm25_idx')
LIMIT 10;
```

Output (correct, matches standalone scoring):
```
NOTICE:  BM25 index scan: tid=(181968,4), BM25_score=-23.0376    -- doc 3906880
...
 passage_id |  score
------------+---------
    7759479 | 27.9390
    7759478 | 23.5952
    3906880 | 23.0376   ← rank 3, correct
    7759484 | 21.5800
```

### Query B: row_number() OVER (...) blocks LIMIT pushdown → BMW gets K = default_limit = 1000

```sql
SELECT row_number() OVER () as rank, passage_id,
       -(passage_text <@> to_bm25query('3 ways a log can move when bucking on side hill', 'msmarco_bm25_idx'))::float8 as score
FROM msmarco_passages
ORDER BY passage_text <@> to_bm25query('3 ways a log can move when bucking on side hill', 'msmarco_bm25_idx')
LIMIT 10;
```

Output (wrong: BMW reports a lower score for tid=(181968,4)):
```
NOTICE:  BM25 index scan: tid=(181968,4), BM25_score=-20.6486    -- doc 3906880, **2.39 too low**
...
 rank | passage_id | score
------+------------+---------
    1 |    7759479 | 27.9390
    2 |    7759478 | 23.5952
    3 |    7759484 | 21.5800     ← wrong order
    4 |    3906880 | 23.0376     ← was rank 3 in Query A
```

(The score *column* shows 23.0376 because it's re-evaluated standalone at the WindowAgg level. The score that BMW returned to the index scan was -20.6486, which is the order BMW returned the rows in. Standalone score and BMW score disagree.)

### Confirming K is the cause

```sql
SET pg_textsearch.default_limit = 10;
-- Query B now gives the SAME result as Query A.
```

## BMW stats

```
Query A (K=10):   results=10,   segments=6613 docs,   blocks scanned=197975, skipped=99242 (33.4%)
Query B (K=1000): results=1000, segments=76876 docs,  blocks scanned=442729, skipped=168660 (27.6%)
```

K=1000 keeps the heap threshold lower throughout the scan, leading to more docs scored and more pivots evaluated. Somewhere in this longer traversal, scores diverge from standalone for specific docs.

## Where it likely is

The pivot algorithm in `score_segment_multi_term_bmw` / `score_pivot_document` (src/scoring/bmw.c). My current hypothesis:

The pivot mechanism scores a doc D using contributions from terms in positions `[0..pivot_len-1]` whose `cur_doc_id == D`. For a doc whose contributions come from terms that aren't all in the pivot range at the time D is scored (because some term's iterator has skip-advanced past D via `seek_term_to_doc` to a doc > D, but D was in that term's posting list), the contribution is lost.

This is masked at small K because the heap threshold rises fast and BMW prunes aggressively — fewer pivots evaluated, less chance of mis-aligning. At large K, threshold stays low, more pivots get scored, and the alignment gap shows up.

I'm not 100% sure on the exact mechanism — needs more careful code reading and instrumentation.

## Why this matters

- **Validation correctness**: `validate_queries.sql` (this repo's benchmark watchdog) uses a query shape that prevents LIMIT pushdown to the index scan, defaulting K to 1000. This caused queries 1267, 130, 1517, 222, 9351, 9340, and probably others to falsely fail validation on the concurrent-INSERT path, including across recent benchmark runs. Many runs of #360 "flakiness" trace back here.
- **Application correctness**: any application that asks for top-N via the BM25 index with the same SQL pattern (window function above the LIMIT) will get scores that disagree between BMW and standalone. Ranking is also affected — the ORDER BY output is what BMW returns, not what standalone would give.
- **The bug is silent**: BMW returns plausible-looking scores and doesn't error out. It just reports lower scores than truth for a subset of docs.

## Workaround

Use SQL shapes that allow LIMIT pushdown to the index scan: put LIMIT in the same SELECT block as ORDER BY, before wrapping in row_number()/window functions. (This is what the immediate PR for validate_queries.sql does.)

## Followup

Real fix is at the BMW algorithm level — ensure score correctness for any doc that enters the heap, regardless of K.

Discovered while investigating #361 / #363 from PR #360 and #364.

tjgreen42:
Fixes #365.

## Bug

`score_segment_multi_term_bmw` decided whether to short-circuit a block via `block_max_skip_advance` using only the pivot terms' block-level upper bound:

```c
block_upper = compute_block_max_at_pivot(terms, pivot_len);
if (block_upper <= threshold) {
    block_max_skip_advance(...);   /* advances one pivot term past current block */
    continue;
}
```

`block_max_skip_advance` then advances one of the pivot terms past its current block. But docs in that block may also appear in *non-pivot* terms' posting lists — when a different pivot iteration later picks those docs up, `score_pivot_document` only iterates `[0..pivot_len)`, so the contribution from the term that was already advanced past is silently missing from the doc's reported score.

The looser the heap threshold (large K), the more frequently the old check fires for blocks whose docs are still top-K candidates once non-pivot contributions are included — exactly the K-dependent score under-reporting seen on the MS MARCO benchmark.

## Fix

Tighten the safe-skip condition to:

```
block_upper(pivot) + sum(max_score(non-pivot)) <= threshold
```

This is a true upper bound on any score we could later compute for a doc in the skipped block, no matter which pivot picks it up.

## Verification

**Local repro on MS MARCO (8.8M docs, pg17, concurrent inserts) — query 1267 doc 3906880**:

| K | Before | After |
|---|--------|-------|
| 10 | -23.0376 ✓ | -23.0376 ✓ |
| 1000 | -20.6486 ✗ (off by 2.39) | -23.0376 ✓ |

**Regression tests**: `bmw`, `bmw_skip_advance`, `wand`, `scoring1..6` all pass.

(`max_shared_memory` fails locally for me both with and without this patch — unrelated, looks like a DSA-state issue on my long-running pg17 cluster; will verify in CI.)

This is the v1.2 blocker called out in #365.

### DIFF
--- a/src/scoring/bmw.c
+++ b/src/scoring/bmw.c
@@ -1528,14 +1528,49 @@ score_segment_multi_term_bmw(
 		/*
 		 * Step 3: Block-max refinement.
 		 * Check if block-level upper bound still beats threshold.
+		 *
+		 * Correctness note (#365): block_max_skip_advance() advances
+		 * a pivot term past its current block via seek_term_to_doc.
+		 * Docs in the skipped block that exist in *non-pivot* terms'
+		 * posting lists too can still be top-K candidates -- their
+		 * total true score includes non-pivot term contributions
+		 * that the pivot's block_upper alone doesn't capture.
+		 *
+		 * The safe-skip condition is therefore:
+		 *   block_upper(pivot) + max_score_sum(non-pivot) <= threshold
+		 * (an upper bound on a hypothetical doc's score using
+		 * pivot terms' actual block-max plus non-pivot terms' global
+		 * max). Only then can we skip without losing a top-K
+		 * candidate -- or under-scoring one we later see via another
+		 * pivot, after a relevant term already jumped over it.
+		 *
+		 * Previous version used block_upper alone, which let
+		 * block_max_skip_advance jump a term past a doc that other
+		 * pivot iterations would later pick up with one term missing
+		 * from its score sum. Manifests at large K (low heap
+		 * threshold) where the looser check fires for blocks whose
+		 * docs are still top-K when non-pivot contributions count.
 		 */
 		block_upper = compute_block_max_at_pivot(terms, pivot_len);
 
-		if (block_upper <= threshold)
 		{
-			block_max_skip_advance(
-					terms, term_count, pivot_len, &active_count, stats);
-			continue;
+			float4 non_pivot_max = 0.0f;
+			int	   np;
+
+			for (np = pivot_len; np < term_count; np++)
+			{
+				TpTermState *ts = terms[np];
+				if (!ts->found || ts->iter.finished)
+					continue;
+				non_pivot_max += ts->max_score;
+			}
+
+			if ((block_upper + non_pivot_max) <= threshold)
+			{
+				block_max_skip_advance(
+						terms, term_count, pivot_len, &active_count, stats);
+				continue;
+			}
 		}
 
 		if (stats)


tjgreen42:
## Benchmark run status

Manually-triggered benchmark on this branch: https://github.com/timescale/pg_textsearch/actions/runs/25713784697

✅ **MS MARCO validation** — was failing on main, now passes with this fix (the #365 reproducer query 1267 doc 3906880 now scores correctly at all K).

⚠️ **Wikipedia validation** — fails with score mismatch on 1/80 queries and **0/80 top-10 doc-set matches**. After investigating: this is a pre-existing dataset/validator rot (filed as #368), **not a regression from this fix**.

Briefly: before #360, the MS MARCO validator aborted on an `ambiguous query_id` error before its `DROP FUNCTION` cleanup, leaving its `validate_single_query` overload in `bench_test`. The Wikipedia validator then collided with that overload at `LATERAL validate_single_query(int, text)` and aborted too — but the pre-#360 watchdog lacked `set -o pipefail` and the "VALIDATION PASSED" marker check, so both silently "passed". The hardening in #360 plus the MS MARCO column fix unmasks the Wikipedia GT/index mismatch that's been latent for a while.

See #368 for full root-cause analysis and proposed unblock paths.

CI: all 17 standard checks pass (sanitizers pg17/18, scoring, bmw, wand, bmw_skip_advance, performance, format, etc.).

PR #375, opened 2026-05-14
Replace shared-memory memtable with on-disk paged structure (#374)

### LINKED ISSUE #374: Split SPILL / MERGE_LINKAGE: emit metapage / segment-header page changes via GenericXLog
## Summary

Proposal to split `XLOG_TP_SPILL` (and `XLOG_TP_MERGE_LINKAGE`) so the on-disk page mutations they carry can be replayed without loading `pg_textsearch.so`. This makes recovery in shared-state-less environments (e.g. PostgreSQL forks that run a standalone WAL-replay helper to reconstruct individual pages) work correctly without dlopening the extension.

## Background

Today the custom rmgr (`TP_RMGR_ID = 149`) handles three record types:

- `XLOG_TP_INSERT_TERMS` – purely memtable: re-applies a doc's term list to the in-shared-memory memtable. Has no block refs.
- `XLOG_TP_SPILL` – bundles two distinct effects:
  1. **Page-level**: updates the metapage's `level_heads[0]` / `level_counts[0]` (and optionally the new segment header's `next_segment`) — applied via `XLogReadBufferForRedo` and gated only on `BLK_NEEDS_REDO`.
  2. **Memtable-level**: clears the in-shared-memory memtable — gated on `registry_hit` (per-index shared state must exist).
- `XLOG_TP_MERGE_LINKAGE` – same pattern: metapage linkage update (always applied), plus atomic-counter shrink (registry-gated).

The bulk segment data itself already replicates via `GenericXLog` from `segment.c`, so it doesn't go through the custom rmgr.

## Problem

Some PostgreSQL deployments run a standalone "WAL-redo" helper process that reconstructs a target page by replaying only the WAL records that touch it. That helper has no `shared_preload_libraries`, no shared memory, no per-index registry, and no memtable — but it does need to apply the page mutations carried in `SPILL` / `MERGE_LINKAGE` records to faithfully reconstruct the metapage or a segment header page.

Because the custom rmgr ID is unknown to such a helper unless `pg_textsearch.so` is explicitly dlopened first, the helper has to load the extension purely to keep `RmgrTable[149].rm_redo` non-NULL. The `_PG_init` walredo fast-path (`if (am_wal_redo_postgres)`) already short-circuits the rest of init, but the load itself is still required.

This is a workable but awkward coupling: an extension that exists to provide in-memory acceleration ends up entangled with the recovery path of relations that, from a page-format perspective, look like ordinary index relations.

## Proposal

Separate the two concerns at the WAL-emission boundary:

1. **Page-level effects** (metapage chain-link updates, segment header `next_segment` updates) → emit as standard `GenericXLog` records. These can already be applied by *any* PostgreSQL replay environment without loading any extension, since `GenericXLog` is core PG.

2. **Shared-state effects** (memtable clear, atomic counter shrink) → keep as TP_RMGR records, but make them *page-less* (no `XLogRegisterBuffer`, no block refs).

After the split, the TP_RMGR records carry only in-memory side effects. A backend with no per-index registry can correctly handle these by no-op'ing, and a recovery helper that never has shared state can be told (or naturally infer) that TP_RMGR records have nothing relevant to a specific page reconstruction.

## Benefits

- Recovery environments without shared state can replay any page modification without dlopening `pg_textsearch.so`. The walredo fast-path in `_PG_init` becomes unnecessary.
- Cleaner separation of concerns at the WAL layer: page-format invariants ride on a core, well-understood facility (`GenericXLog`); custom rmgr records carry only the extension-specific in-memory state.
- Smaller surface area for the `am_wal_redo_postgres` fast-path; arguably the conditional can be removed entirely.
- Easier to reason about replay semantics in mixed environments (PITR + missing extension, crash recovery before extension finishes init, etc.).

## Considerations

- The split changes the on-disk WAL format for `pg_textsearch`. Would need an upgrade/compat strategy — likely tied to a `pg_textsearch` version bump and require a clean restart through the boundary version.
- `GenericXLog` is slightly more verbose on the wire than the current packed `xl_tp_spill` / `xl_tp_merge_linkage` headers. For metapage linkage updates this is a few bytes of overhead per record, probably negligible compared to segment data volume.
- Locking semantics need to be preserved: the LW_EXCLUSIVE held during `MERGE_LINKAGE` redo to block in-flight scans is important. After the split, that lock could be acquired in the TP_RMGR record's redo (which would still exist, just without page refs).

## Alternatives Considered

- **Per-record skip flag in walredo**: have the recovery environment know which rmids are "safe to skip" when no block refs match. Doesn't generalize, and silently skipping is dangerous because a record might still touch the target page via blockless side effects.
- **Make `pg_textsearch.so` part of the recovery image's standard base**: works, but couples extension distribution to recovery-helper distribution.

The WAL-record split feels like the cleanest long-term direction.

Happy to put up a PR if there's interest in this approach.

tjgreen42:
Replaces the shared-memory memtable with an on-disk paged structure, removing the custom WAL resource manager and the docid recovery pages it required.

Closes #374. (Supersedes the design proposed there, which kept an in-memory memtable behind custom-but-page-less WAL records.)

## Motivation

The original L0 layer was a shared-memory inverted index (DSA-backed dshash tables) replayed on standbys via a custom rmgr (ID 149). This worked on a regular streaming standby — extension loaded, rmgr registered, redo applies to the in-shmem memtable — but it broke PostgreSQL's stateless WAL-redo helper used for single-page reconstruction: that environment has no shared memory and cannot load `pg_textsearch.so`, so `INSERT_TERMS` redo is undefined and `SPILL`/`MERGE_LINKAGE` are brittle.

Rather than re-engineer custom WAL records to be page-only, this PR removes the in-memory memtable entirely. The memtable becomes a chain of pages in the index relation itself, mutated under buffer locks, WAL-logged via `GenericXLog`. Walredo and any other stateless replay environment reconstructs every page without loading the extension.

## What changes

- **Memtable storage.** A chain of pages rooted at the metapage's new `(memtable_head_blkno, memtable_tail_blkno)` fields. Each page has a magic header (`TP_MEMTABLE_PAGE_MAGIC`) and a sequence of variable-length doc records (`ctid, doc_length, vector_len, vector_bytes`). Oversized `TpVector` payloads are stored as a FRAGMENT head + N CONTINUATION pages, reassembled by the reader.
- **Write path.** `tp_add_document_terms` builds a `TpVector` and calls `tp_memtable_append`, which extends or appends to the tail page under a single `GenericXLog` record. Per-index `LW_SHARED` is held for the txn; the tail buffer is taken `EXCLUSIVE` only while the record is being written.
- **Read path.** `chain_source` walks the chain under the per-index `LW_SHARED`. Scoring and query-time `doc_freq` paths take a `TpDataSource *` so they consult the chain instead of dshash.
- **Spill.** `tp_do_spill` extracts the chain via `chain_source`, builds an L0 segment via `tp_write_segment`, then one `GenericXLog` over `{metapage, new segment header}` publishes the result. Held under `LW_EXCLUSIVE`.
- **Custom rmgr deleted.** `src/replication/rmgr.c` is gone, the call to `tp_register_rmgr()` in `_PG_init` is removed, and the `INSERT_TERMS` / `SPILL` / `MERGE_LINKAGE` record types no longer exist. Slot 149 is intentionally retired.
- **`pg_textsearch.memory_limit` and `bm25_memory_usage()` removed.** The soft-limit byte-accounting was DSA-specific and not meaningful for the page-backed layer. Replaced by a chain-page-count trigger `pg_textsearch.memtable_pages_threshold` (default 64). Postgres' shared-buffer pressure is the real cap on memtable size now; an oversized chain just gets paged out by the buffer manager like any other relation.
- **`MarkGUCPrefixReserved("pg_textsearch")`** added to `_PG_init` so stale `postgresql.conf` entries for the removed GUCs are flagged at server start instead of being silently ignored.
- **PARALLEL UNSAFE** marker added to `bm25_text_bm25query_score` and `bm25_textarray_bm25query_score`: the chain source needs a stable per-index lock for the lifetime of the query.

## Spec

See [`docs/memtable_v2.md`](docs/memtable_v2.md) for the on-disk page format, write/read/spill flow, concurrency contract, and WAL replay story. Replaces the obsolete `docs/replication/CORRECTNESS.md`.

### Bulk path

| Path                        | main 1.2.x | this PR  | delta    |
|-----------------------------|------------|----------|----------|
| 50k bulk INSERT             | 2.48s      | 2.20s    | **-11%** |
| 10k INSERT (post-spill)     | 232ms      | 204ms    | **-12%** |

### Read latency vs. memtable chain size

By design, read latency scales with the size of the in-flight memtable chain (see [`docs/memtable_v2.md`](docs/memtable_v2.md)).  How full the chain is allowed to get is a per-deployment tradeoff governed by `pg_textsearch.memtable_pages_threshold` (default 64).

| Memtable size                 | Chain pages | this PR  |
|-------------------------------|-------------|----------|
| 0 docs (segments-only)        | 0           | 2–5 ms   |
| 2k docs (default threshold)   | 64          | 3.2 ms   |
| 10k docs                      | 255         | 14 ms    |
| 30k docs                      | 960         | 59 ms    |

GH `Benchmarks` workflow run for full numbers vs. the most recent successful `main` scheduled run is in the comments below.

## Upgrade

`TP_METAPAGE_VERSION` is bumped 6 → 7. **v6 metapages fail at open with a clear `REINDEX INDEX` error**; there is no dual-format support. The `pg_textsearch--1.2.0--1.3.0-dev.sql` upgrade script handles the SQL-level side (drops the removed `bm25_memory_usage` function, marks scoring functions PARALLEL UNSAFE), but operators must `REINDEX` each `bm25` index to migrate v6 metapages to v7. The upgrade-tests CI matrix exercises this path for v1.2.0 (added in Batch 4 of post-review fixes).

## Out of scope / follow-ups

- #377 — Drop `shared_preload_libraries` requirement (subsumes registry teardown / `TpMemtable` deletion)
- #378 — Release rmgr ID 149 from the PostgreSQL wiki
- #379 — Reclaim orphaned chain pages after spill
- #380 — Standby-safe horizon for displaced segment pages during merge (predates this PR)
- #381 — Reduce WAL amplification for FRAGMENT chain-page writes
- #382 — Tail-page contention on high-concurrency single-term inserts
- #383 — Re-establish read compatibility for v6 metapages (avoid forced REINDEX on upgrade)

## Testing

- `make installcheck`: **64/64 green**.
- `make format-check`: clean (clang-format 21.1.8).
- `bash test/scripts/wal_audit.sh`: all 3 tests pass — verifies zero custom rmgr records in any LSN window, including across a SIGKILL + restart crash window.
- `bash test/scripts/shutdown_spill.sh`: pass.
- `bash test/scripts/docid_chain_recovery.sh`: pass (test now exercises the new memtable chain; file rename deferred to a follow-up).
- `pg_waldump` on the regression's traffic: zero rmgr ID 149 records.


tjgreen42:
# memtable-v2 update-heavy benchmark + profile (PR #375)

## Setup
- PG18.3 on WSL2, 16 GB shared_buffers=4GB, max_wal_size=1GB, synchronous_commit=on
- pg_textsearch 1.3.0-dev at `118ff1ef` (`memtable-v2` branch, HEAD)
- Dataset: MS MARCO passages, first 200k loaded into staging
- pgbench drives one-row INSERTs from staging → indexed table

## Headline numbers (default GUCs, msmarco passage workload)

| Workload                                | TPS    | avg ms  | p99-ish (stddev) |
|-----------------------------------------|--------|---------|------------------|
| INSERT, c=1                             | 693    | 1.44    | σ=1.13           |
| INSERT, c=4                             | 1951   | 2.04    | σ=4.7            |
| INSERT, c=8                             | 3123   | 2.55    | σ=7.9            |
| READ, c=4 (post-spill, segments only)   | 1572   | 2.53    | σ=0.95           |
| READ + INSERT, c=4+4 (default threshold)| 442    | 8.97    | σ=2.7            |

Insert throughput scales ~4.5× from 1 → 8 clients (sub-linear; expected — tail-page latch + WAL fsync contention).

Read-while-writing at default thresholds: **3.5× slower** than read-only. Cause analyzed below.

## Where the time goes

### Wait events during c=8 INSERT (180 samples, 100ms apart)

```
179 LWLock|WALWrite      ← WAL writer lock (PG core)
 59 CPU|CPU
 56 LWLock|extension     ← relation extension lock (PG core)
 38 IO|WalSync           ← fsync on commit
  4 LWLock|BufferContent ← buffer content latch
```

- No pg_textsearch-named LWLock appears in waits. **The per-index
  LWLock + tail-page buffer latch are not the bottleneck.**
- Dominant insert cost is WAL fsync + relation extension — both PG
  core, not extension-specific.

### Read-side profile (gdb stack sampling, 20 samples)

Driver: 600 BM25 queries on a 639-page (≈20k doc) memtable chain.
Single query: ~106 ms.

Stack frame distribution (every sample is in the same call path):

```
tp_execute_scoring_query  (scan.c:393)
  → tp_memtable_search  (scan.c:101)
    → tp_score_documents  (bm25.c:168)
      → tp_memtable_chain_source_create  (chain_source.c:811)
        → walk_chain  (chain_source.c:583)
          → ingest_terms  (chain_source.c:293)
            → lookup_or_create_term  (chain_source.c:163)
              → dynahash hash_search        ← ~60% of CPU
            → term_entry_append  (chain_source.c:203)
              → AllocSetAlloc / repalloc    ← ~15%
```

Top leaf frames (by count):

| Frame                                          | Samples |
|------------------------------------------------|---------|
| dynahash.c:1019-1025 (hash_search_with_hash_value) | 11   |
| strcmp_avx2 (key compare)                      | 2       |
| hashfn.c:216 (hash compute)                    | 1       |
| AllocSetAlloc / MemoryChunkSetHdrMask          | 1       |
| (other dynahash internals)                     | 5       |

**Verdict.** Time is spent in dynahash term lookups during memtable
chain ingestion. The current chain source walks every record and
hashes every term into a transient per-query HTAB. For a memtable
of M chain pages and queries with N unique terms per doc (≈30 for
msmarco), the per-query cost is **O(M × records-per-page × N)**
hash operations.

At default `pg_textsearch.memtable_pages_threshold = 64` pages this
is ~60k hashes/query → about 10-15 ms read overhead on a "full"
chain, matching Phase 10's reported 1.5→15 ms regression.

## Update-heavy scenario (raised spill threshold)

To see what update-heavy tuning looks like, we raised
`pg_textsearch.memtable_pages_threshold` to 100000 (effectively
disabling auto-spill for this run):

| Memtable chain depth | Read latency (single query) |
|----------------------|-----------------------------|
| 0 pages   (segments-only)              | 2-5 ms        |
| 639 pages (20k docs in memtable)       | 106 ms        |
| 961 pages (30k docs in memtable)       | 162-187 ms    |

Per-record cost is ~6 μs (dynahash lookup + memcpy + repalloc).
This is linear in chain size by construction (Decision 1: doc-record
log shape, no per-term skip data).

WAL bytes per insert (chain-page memtable):
- c=1: 1378 B
- c=4: 1642 B
- c=8: 1825 B

Reasonable (≈heap tuple + bm25vector encode + 1 chain page diff).
Sub-linear growth with concurrency = FPI hits when tail page crosses
checkpoint boundaries.

## Read+insert contention sample (mixed phase, 30k-doc memtable)

```
                                count   wait
                                4936    (idle, between txns)
                                 311    CPU|CPU
                                  59    LWLock|WALWrite
                                  28    IO|WalSync
                                   1    LWLock|BufferContent  ← single sample
                                   1    Client|ClientRead
```

Read backends are **never** observed waiting on a pg_textsearch
LWLock. Their wait_event is consistently `CPU|CPU` — they are CPU-bound
in the chain_source walk, not blocked on inserters. The single
`LWLock|BufferContent` sample is the buffer-content lock on a chain
page (reader SHARED vs writer EXCL on the tail).

One sample also caught a read backend blocked in
`tp_acquire_index_lock` → `LWLockAcquire(LW_SHARED)` → futex wait.
This happens when a spill holds `LW_EXCLUSIVE`; the wait stops as
soon as spill completes. At default thresholds spills are fast (≈64
pages → ~10ms), so this contention is bounded.

## Conclusions

1. **Insert path scales well.** TPS scales 4.5× from c=1 to c=8.
   Limiting factors are PG-core WAL fsync + relation extension —
   not the memtable design.

2. **Read latency on a built-up memtable is linear in chain size, as
   predicted by the plan.** With default `memtable_pages_threshold=64`
   the chain stays small enough that this is the documented 1.5→15 ms
   delta. Tuning the threshold higher (for write-heavy workloads)
   pays linearly: 64 → 1000 pages ≈ 10 ms → 160 ms.

3. **No surprising hotspot.** The hot path (>95% of samples) is
   `chain_source.c::lookup_or_create_term` calling dynahash.
   This is the expected cost: every query re-builds a per-term
   HTAB from scratch by walking every record.

4. **Concurrency contract works as designed.** No buffer-lock or
   LWLock starvation observed. Reads serialize against spills only
   (LW_EXCLUSIVE), and spills are short.

## Recommended follow-ups (separate issues)

A. **Per-backend chain-source cache (Phase 10 idea, deferred).**
   Cache the deserialised term HTAB between queries on the same
   backend; invalidate on metapage `(tail_blkno, n_records,
   spill_generation)` change. Walk only new records since last
   cache. Eliminates 90%+ of the dynahash cost for read-heavy
   workloads against a growing chain.

B. **Term-table pre-sizing.** Current dynahash starts at ~16 buckets
   and grows. For a 20k-doc memtable, term cardinality is in the
   tens of thousands. Pre-sizing the HTAB at `chain_pages × avg_terms_per_page`
   would eliminate ~50% of rehashes.

C. **Cheaper key.** dynahash key is `const char *` with custom hash
   over bytes. For lexemes (already normalised, ≤256 bytes), a
   simhash + linear probe over a flat array of `(hash, term_offset,
   posting_idx)` triples would be 2-4× faster.

D. **Inverted-shape memtable (Plan A2).** The plan was explicit
   that A1 was chosen for simplicity; if read latency on growing
   memtables becomes a frequent ops complaint, A2 (paged inverted
   index in the memtable, with WAL via GenericXLog) eliminates the
   per-record term walk entirely. Significant work (≈2500 LOC).

(A) is the single highest-leverage optimisation against the user's
specific concern of "query performance for update-heavy workloads"
— it makes the dominant cost amortised, not per-query.

## Raw data

- /tmp/bench_v2/results/pgbench_c{1,4,8}_5k.txt — per-concurrency runs
- /tmp/bench_v2/results/pgbench_mixed_*.txt — mixed read+insert
- /tmp/bench_v2/results/pgss_*.txt — pg_stat_statements snapshots
- /tmp/bench_v2/results/waits_*.txt — wait-event samples
- /tmp/bench_v2/stacks2.txt — gdb stack samples (20)
- /tmp/bench_v2/results/focused.log — full focused-bench transcript

tjgreen42:
## Read-path optimization: query-term HTAB filter (commit `f1e82521`)

Per Todd's observation that we were building a term HTAB containing
*every* unique lexeme in the memtable when queries only consume a
handful of them, `tp_memtable_chain_source_create` now accepts an
optional `query_terms[] / query_term_count` and, in that mode:

1. Pre-populates the term HTAB with exactly those terms (pre-sized
   to N buckets, no dynahash bucket-split rehashes).
2. Switches `ingest_terms`'s per-record loop from `HASH_ENTER` to
   `HASH_FIND` — non-query lexemes are skipped before any posting
   buffer is appended.

Spill (`tp_do_spill` via `tp_memtable_chain_source_extract`) and
corpus-totals callers (`to_bm25query`, fresh-backend bootstrap) keep
the unfiltered path; only the BMW scoring entry point
(`tp_score_documents`) opts in.

### Measured impact

30k-doc / 960-page chain (msmarco passages, default GUCs except
`memtable_pages_threshold = 100000` to suppress auto-spill):

| Query           | Before  | After  | Speedup |
|-----------------|---------|--------|---------|
| `quantum` (1 term) | 60 ms  | 60 ms | parity  |
| `cancer treatment` (2 terms) | 162-187 ms | 67 ms | **2.7-3.1×** |

The 1-term path was already CPU-bound on TpVector decode + doc-length
HTAB build (not term HTAB churn), so it's unchanged. The 2-term path is
the one this commit unblocks.

### Profile confirmation

Before: `gdb -batch -ex "bt 20"` sampling, 100% of stacks in:
```
walk_chain → ingest_terms → lookup_or_create_term
  → dynahash hash_search HASH_ENTER (~60% CPU)
  → term_entry_append → repalloc (~15% CPU)
```

After: `hash_search` HASH_ENTER is gone from the hot path; remaining
samples are inside the `TpVector` byte-walk itself (`ingest_terms`
lines 313/329, the lexeme-iteration code).

### Why not 30×?

The per-record cost remaining is the TpVector decode (varint walk +
per-term length read) and the 30k-entry doc-length HTAB build — both
inherent to the chain shape. Future wins would come from:

- Bloom-filter prefilter on the page header (skip pages with no query
  term hit).
- Smaller / packed doc-length representation (e.g., sorted array on
  the page).
- A2-style inverted memtable (out of scope; see plan).

### Verification

- `make installcheck`: 64/64 tests pass.
- `make format-check`: clean (clang-format 21.1.8).
- Commit: `f1e82521`, pushed to `memtable-v2`.

tjgreen42:
## TpVector decode: skip redundant validate + fuse decode/advance (commit `5bd3a273`)

Profiling after `f1e82521` (query-term HTAB filter) put `dynahash`
out of the top of the stack and exposed the next layer: the
TpVector entry-walk inside `ingest_terms`. Two small wins:

### 1) Skip `tpvector_validate_v2_buffer` on the chain read path

`ingest_terms` was calling `tpvector_validate_v2_buffer` on every
record, which walks each entry's varints purely to bounds-check —
i.e., the same work the ingest pass itself does. But the bytes
come from chain pages we wrote ourselves via
`tp_add_document_terms` → `create_tpvector_from_strings`; the v2
buffer is assembled in-process and never crosses a trust boundary
on the way to the reader. PG buffer-CRC handles disk-level
corruption; structural corruption of our own write would point at
our bug, not untrusted input.

Gated behind `USE_ASSERT_CHECKING`: debug builds keep the
cross-check, production builds skip the redundant varint walk.
Operator entry-point validators (`tpvector_canonicalize` for
recv/out/eq) and the spill path are unchanged — they remain the
trust boundary for externally-sourced bytes.

### 2) Fuse entry decode + advance into one varint pass

The legacy iterator pattern

```c
tpvector_entry_decode(e, &v);     // decode freq + lex_len
... use v ...
e = get_tpvector_next_entry(e);   // re-decode both varints to advance
```

re-decoded both varints just to compute the next cursor, even
though the first decode already produced `v.lexeme + v.lexeme_len`.
For a 30k-record × 30-term workload that's ~1.8M unnecessary
varint decode loops per query.

New `tpvector_entry_decode_advance(entry, &v) → next` returns the
cursor the varint loop already produced. Updated the three
in-loop callers — `chain_source.c:ingest_terms`,
`memtable/scan.c` query-vector parse, `access/build.c:tp_insert`.
Legacy `tpvector_entry_decode` / `get_tpvector_next_entry` stay
for callers that need either operation independently (output
formatting, equality).

### Measured impact

30k-doc / 960-page chain (msmarco passages, default GUCs except
`memtable_pages_threshold = 100000`):

| Query                        | After `f1e82521` | After `5bd3a273` | Δ      |
|------------------------------|------------------|------------------|--------|
| `quantum` (1 term)           | 60 ms            | **41 ms**        | −32%   |
| `cancer treatment` (2 terms) | 67 ms            | **59 ms**        | −12%   |
| 4 readers solo (avg)         | 61 ms            | **47 ms**        | −23%   |

The 1-term path benefits more here because the query-term filter
didn't change its work profile (only 1 term in HTAB either way);
the TpVector decode was its dominant cost.

### Cumulative speedup from PR baseline (`118ff1ef`):

| Query                        | Before  | After `5bd3a273` | Speedup |
|------------------------------|---------|------------------|---------|
| `quantum` (1 term)           | 60 ms   | 41 ms            | 1.5×    |
| `cancer treatment` (2 terms) | 187 ms  | 59 ms            | **3.2×**|

### Verification
- `make installcheck`: 64/64 pass.
- `make format-check`: clean (clang-format 21.1.8).
- Commit `5bd3a273`, pushed.

tjgreen42:
## GH Benchmarks: memtable-v2 vs main

GH run [25890704268](https://github.com/timescale/pg_textsearch/actions/runs/25890704268) (memtable-v2) — all 4 bench jobs succeeded (incl. `insert-benchmark`, which is broken on `main` per #376). Comparison baseline: last clean main scheduled run [25420768617](https://github.com/timescale/pg_textsearch/actions/runs/25420768617) (2026-05-06).

### Bulk-index path (full-benchmark) — as expected, essentially unchanged

| Dataset | Index build Δ | Index size Δ | Query weighted-p50 Δ |
|---|---:|---:|---:|
| cranfield (1.4K docs) | −15.2% | ±0.0% | n/a |
| wikipedia (100K docs) | −12.1% | ±0.0% | −7.7% |
| msmarco (8.8M docs) | −2.7% | ±0.0% | **+17.8%** |

Index *size* is byte-identical for the bulk path on all three datasets (memtable is fully drained by the time CREATE INDEX returns) — good signal that segment format is unchanged. Build time is slightly faster on cranfield/wikipedia, ~flat on msmarco.

Query latency on `full-benchmark`:
- wikipedia: flat / slightly faster across all token-count buckets
- msmarco: 1-tok faster (-13%); 2- through 8-tok bucket p50 slower by 8–27%, weighted-p50 +17.8%

The msmarco read-path regression is the residual from the chain-source path. The 30k-doc local-bench improvement we observed (187 → 59 ms = 3.2×) shows up best when the memtable is large. In `full-benchmark`, the memtable is drained at index-build time, so what's left is per-segment scoring; the +17% is most likely the cumulative cost of the metapage abstraction layer + chain-source still being on the read hot path even for pure-segment queries.

### Insert / concurrent-insert paths — significant regressions (expected)

This is the trade-off you anticipated when we switched to the on-disk memtable. The chain-page layout costs more to scan than the old DSA hash table.

**msmarco (8.8M passages):**

| metric | memtable-v2 | main 5/06 | Δ |
|---|---:|---:|---:|
| Serial insert time | 900.6 s | 727.4 s | **+23.8%** |
| Concurrent insert time | 1676.6 s | 1513.4 s | +10.8% |
| Index size after serial insert | 4557 MB | 2196 MB | **+107.5%** |
| Weighted p50 (serial-insert) | 11.06 ms | 7.27 ms | **+52.1%** |
| Weighted p50 (concurrent) | 6.33 ms | 4.83 ms | +31.1% |
| 8-tok p50 (serial-insert) | 30.8 ms | 23.4 ms | +31.5% |

**wikipedia (100K articles):**

| metric | memtable-v2 | main 5/06 | Δ |
|---|---:|---:|---:|
| Serial insert time | 25.6 s | 22.0 s | +16.6% |
| Concurrent insert time | 24.3 s | 19.1 s | +27.3% |
| Index size after serial insert | 122 MB | 50 MB | **+144.8%** |
| Weighted p50 (serial-insert) | 0.94 ms | 0.25 ms | +276% |
| Weighted p50 (concurrent) | 3.43 ms | 0.79 ms | +334% |

Wikipedia query-latency regressions are large in %-terms but small in absolute terms (sub-ms → low-ms p50s).

The +100–150% index-size deltas reflect chain pages on disk + segments not yet compacted, since `insert-benchmark` measures the post-insert state without running CREATE INDEX. With default tuning, the workload didn't trigger enough spills to keep the memtable small.

### MSMARCO vacuum index size

| metric | memtable-v2 | main 5/06 | Δ |
|---|---:|---:|---:|
| msmarco_vacuum index size | 2548 MB | 1301 MB | **+95.9%** |

Worth a separate look — post-VACUUM index size doubled. This is downstream of the same insert-path size delta, but I want to confirm the alive-bitset path still reclaims as expected on the new memtable layout.

### Takeaways

1. **Bulk-index path is unaffected** — index format unchanged, build times marginally faster, msmarco read latency +17% (the chain-source residual).
2. **Insert-path query/space hit is real** (msmarco weighted p50 +52%, wikipedia +275%, index size roughly 2×) — this is the trade-off we accepted when we replaced the DSA memtable with a write-optimized paged layout.
3. **Concurrent insert paths run clean on memtable-v2** — they're broken on `main` (#376), so this PR also incidentally fixes the nightly bench baseline.

The local read-path optimizations (commits `f1e8252` and `5bd3a27`) brought the **memtable-only** read cost from 187 ms → 59 ms on 30k-doc / 960-page scans. They land their biggest wins on workloads that keep the memtable large between flushes; the CI insert benches end up with a relatively cold memtable that's mostly already on disk pages.

Artefacts: [v2 run 25890704268](https://github.com/timescale/pg_textsearch/actions/runs/25890704268) · [main baseline 25420768617](https://github.com/timescale/pg_textsearch/actions/runs/25420768617).

tjgreen42:
## Post-merge nightly bench: 🟢 all green

Manually-dispatched `Benchmarks` workflow on `main` post-merge: [run 25894810538](https://github.com/timescale/pg_textsearch/actions/runs/25894810538).

| Job | Result |
|---|---|
| `full-benchmark` | ✅ success |
| `baseline-benchmark` | ✅ success |
| `paradedb-benchmark` | ✅ success |
| `insert-benchmark` | ✅ success (incl. MS MARCO concurrent validation) |

The MS MARCO concurrent validation that's been failing on every `main` scheduled run since 2026-05-06 (#376) passed cleanly:

```
VALIDATION PASSED: All 79 queries match within tolerance
  (1 allowlisted known mismatches)
```

The one remaining mismatch is query 158 — the pre-existing stemmer-drift case already allowlisted via #361. Queries 9351 and 9340 did **not** reproduce.

Tomorrow's scheduled run should be green as well; if it isn't, posting back here.

PR #388, opened 2026-05-26
Prevent overflow on 32bit systems.

tdrz:
This adds checks to prevent some pointer overflows on 32bit systems (basic tests now pass on PGlite which is a 32bit).

Unsure if there aren't other places where there is an assumption of this running only on 64bit systems...


tjgreen42:
@tdrz Thanks for the contribution!  Is there actually a realistic use case for hosting full-text search on a 32-bit server?  I'm kind of inclined to not go there, to be honest, since taking it seriously means setting up new CI jobs etc.  But maybe I'm missing something.

tdrz:
> @tdrz Thanks for the contribution! Is there actually a realistic use case for hosting full-text search on a 32-bit server? I'm kind of inclined to not go there, to be honest, since taking it seriously means setting up new CI jobs etc. But maybe I'm missing something.

@tjgreen42 We're building your extension for [PGlite](https://github.com/electric-sql/pglite) (Postgres in wasm) for 32bit.

tdrz:
btw I totally understand if you don't want to support 32bit systems. I will try to keep a fork myself if you decide not to.

tjgreen42:
> btw I totally understand if you don't want to support 32bit systems. I will try to keep a fork myself if you decide not to.

PGlite is an interesting use case.  I'm okay taking unobtrusive fixes like this but I think just want to avoid signing up for a full commitment to 32-bit with CI tests etc.

src/types/vector.c
```diff
@@ -291,7 +291,11 @@ void
 tpvector_entry_decode(const TpVectorEntry *entry, TpVectorEntryView *out)
 {
 	const uint8 *cursor = (const uint8 *)entry;
-	const uint8 *end	= cursor + UINT32_MAX; /* effectively unbounded */
+#if SIZEOF_VOID_P == 4
```
tjgreen42: Can you encapsulate this arithmetic logic inside a macro to keep things tidier (here and below)?

tjgreen42 (changes requested):

tdrz:
> I'm okay taking unobtrusive fixes like this but I think just want to avoid signing up for a full commitment to 32-bit with CI tests etc.

Makes sense, fine with me!

src/types/vector.c
```diff
@@ -342,7 +351,11 @@ tpvector_entry_decode_advance(
 		const TpVectorEntry *entry, TpVectorEntryView *out)
 {
 	const uint8 *cursor = (const uint8 *)entry;
-	const uint8 *end	= cursor + UINT32_MAX; /* effectively unbounded */
+#if SIZEOF_VOID_P == 4
```
tjgreen42: Use the macro here too, no?

tjgreen42 (changes requested):

src/types/vector.c
tdrz: Ups, yes, sorry for that!

tjgreen42 (approved):

PR #392, opened 2026-05-27
memtable: in-memory cache over the on-disk page chain

tjgreen42:
## Summary

Adds an in-memory **memtable cache** layered on top of the on-disk
memtable page chain (issue #374), so query-time scoring reads from a
hot in-process structure instead of re-walking buffered chain pages
on every scan.

The cache is the *only* read path for in-flight (unspilled) documents.
On a miss / unloaded cache the chain is read once, hashed into the
cache under a per-index lock, and subsequent reads in the backend are
zero-buffer-traffic.

## What's in the PR

**New components**

- `src/memtable/cache.{c,h}` — the cache itself: per-index hashtable
  (term → posting list), per-record arena, snapshot-stable apply
  protocol, eviction counters.
- `src/memtable/chain_walker.{c,h}` — single-pass walker over the
  on-disk chain, reusable by the cache loader, the spill path, and
  the scoring chain source.
- `src/memtable/cache_source.{c,h}` — `TpDataSource` adapter so the
  scoring layer asks the cache the same questions it asks segments
  and chain sources.
- Cache-load and cache-aware spill consumption paths in
  `src/memtable/{log,source}.c`.

**Memory governance** (3 tiers, all GUC-driven)

- `pg_textsearch.cache_max_bytes` — soft per-index cap; over the cap,
  the cache stops growing for that index and serves from the chain.
- `pg_textsearch.cache_total_max_bytes` — hard global cap; eviction
  picks the largest cache (LRU within ties) and drops it cleanly,
  taking the registry mutex → victim per-index LWLock (conditional
  EXCL) → victim apply lock → victim cache lock chain so cross-index
  eviction can never deadlock with normal scans.
- Automatic spill of `pg_textsearch.memtable_pages_threshold` chain
  pages (default 64) keeps the chain bounded; cache reload after
  spill is incremental.

**Vestigial-API and atomic cleanup**

The shared-memory `total_docs` / `total_len` atomics on
`TpSharedIndexState` are gone — the metapage holds
`Σ segment.num_docs` / `Σ segment.total_len`, and the chain-source open
adds the in-flight tail. This eliminates a known drift window between
the atomic and the metapage that produced occasional sanitizer-CI
flakes on `unlogged_index.sql`. A handful of unused / phantom
memtable helpers (`tp_sync_metapage_stats`, `tp_doclength_summary`,
`tp_string_table_*`, etc.) are removed.

**Benchmark suite fix**

The insert benchmark used to call `bm25_spill_index()` between the
load and query phases, which emptied the chain so the query phase
ran entirely against segments — making the cache invisible in that
suite. The spill is now in a new `size_report.sql` that runs **after**
the query phase, so query latency reflects the realistic post-load
state (chain + cache + segments). A `benchmark_suite`
workflow_dispatch input also lets a single suite be re-run without
spinning up the others.

## Concurrency contract

- Writers and chain readers continue to hold the per-index LWLock
  SHARED. Spill holds EXCL.
- Cache apply runs under SHARED + a dedicated apply lock; cache
  reads run under cache lock (no chain buffer locks).
- Cross-index eviction takes the registry eviction mutex up-front and
  walks per-index locks conditionally, so a victim that's mid-scan is
  skipped rather than blocked.

## Benchmark results (vs current `main`)

Insert benchmark with the fix above
([run 26598437073](https://github.com/timescale/pg_textsearch/actions/runs/26598437073)) —
query phase exercises chain + cache + segments:

| Dataset                                  | main → feature           |
| ---------------------------------------- | ------------------------ |
| **Cranfield throughput** (1.4k docs)     | 2.30 → 1.56 ms/q (**−32%**) |
| Wikipedia weighted p50 (~150k docs)      | 0.93 → 0.94 ms (flat)     |
| MSMarco weighted p50 (8.8M docs)         | 9.13 → 10.12 ms (+11%)    |
| Insert latency (Cranfield/MSM/Wiki)      | +5% / +5% / +7%          |
| Index size, all datasets                 | byte-identical            |

MSMarco's +11% wp50 is concentrated in bucket-1 single-token queries
(+30% relative on ~3 ms queries); buckets 6–8 are flat. With 8.8M
docs and default 64-page auto-spill, ~0.0007% of MSMarco actually
sits in the chain after load, so the cache pays a small fixed-cost
lookup without much to amortize over.

Concurrent benchmark (queries interleaved with concurrent inserts —
the workload the cache exists for) from the same run:

| Dataset    | wp50 main → feature  |
| ---------- | -------------------- |
| Cranfield (tput)  | 39.54 → 2.51 ms/q (**−94%**) |
| Wikipedia         | 2.90 → 0.90 ms     (**−69%**) |
| MSMarco           | 7.97 → 5.63 ms     (**−29%**) |

Concurrent insert throughput is +10–17% slower across the three
datasets — the cache apply protocol adds work on the write path.
That's a known trade-off and a fair target for follow-up tuning;
it's well within the concurrent-query wins above.

## Tests

- All existing SQL regression tests pass.
- New cache-specific coverage exercises soft/hard cap behavior,
  cross-index eviction, post-spill incremental reload, and the
  metapage-derived corpus-statistics contract.
- Sanitizer build, recovery, concurrency, and segment shell-test
  suites all pass.

### DIFF
--- a/docs/memtable_cache.md
+++ b/docs/memtable_cache.md
@@ -1,0 +1,923 @@
+# Memtable cache — restoring v1 query performance on top of v2 storage
+
+Status: draft (issue: TBD).
+
+## Why
+
+The memtable v2 cutover ([`docs/memtable_v2.md`](memtable_v2.md))
+replaced the v1 shared-memory inverted index with an on-disk
+paged chain of opaque doc records. That change was necessary for
+streaming replication, crash recovery via stock PG replay, and
+single-page WAL-redo compatibility — but it has a real query-time
+cost. Every index scan now opens a `chain_source` that walks
+**every chain page** from `meta.head` to `meta.tail`, decoding
+every doc record, just to materialize the few posting lists the
+query actually needs. With a chain of ~64 pages (the auto-spill
+threshold) that's hundreds of `ReadBuffer` calls per query
+regardless of query selectivity.
+
+The v1 in-memory memtable did not pay that cost: it kept a DSA-
+backed inverted index (`term → posting list`) and a doc-length
+hash table, so a query's posting-list reads were O(query terms)
+not O(chain pages). The cache restores this property without
+changing v2's source of truth.
+
+## Design summary
+
+- The on-disk page chain remains the **source of truth**. All
+  WAL, replication, crash recovery, and standby behavior is
+  unchanged.
+- On top of that chain we maintain a **per-index, DSA-resident
+  cache** of the same content, organized exactly like the v1
+  memtable: dshash `term → TpPostingList`, dshash `ctid →
+  doc_length`, plus rolling corpus totals.
+- The cache is **derived state**: lazily built on first query,
+  incrementally updated by writers, dropped + rebuilt on spill
+  or invalidation. Crashing or losing it is harmless — the next
+  query rebuilds from the chain.
+- An **apply cursor** tracks the next chain record the cache
+  expects to consume. Three paths can advance the cursor —
+  writer hook, reader catchup, cold lazy build — and all three
+  share a single `apply_lock`-serialized helper
+  (`apply_records_up_to`). The cursor IS the commit point;
+  every record is applied to the cache exactly once per
+  generation.
+- The writer hook attaches at the tail of `tp_add_document_
+  terms` while the writer still holds per-index LWLock SHARED.
+  Spill (which needs per-index LWLock EXCL) cannot fire
+  concurrently with cache application, so spill-during-apply is
+  excluded by the existing v2 lock contract.
+
+The cache is the v1 memtable. The chain is the v2 memtable. The
+only conceptual additions are the apply cursor + the shared
+apply protocol that keeps the cache in sync with the chain.
+
+## What we resurrect
+
+The v1 cutover commit (`023c3841`) deleted seven files we want
+back, with surgical adjustments for v2:
+
+| Path                               | LOC | Reuse |
+|------------------------------------|----:|-------|
+| `src/memtable/memtable.h`          |  44 | unchanged surface |
+| `src/memtable/memtable.c`          |  11 | stub, was always tiny |
+| `src/memtable/posting.h`           |  86 | unchanged surface |
+| `src/memtable/posting.c`           | 458 | unchanged |
+| `src/memtable/posting_entry.h`     |  24 | unchanged |
+| `src/memtable/stringtable.h`       |  97 | unchanged surface |
+| `src/memtable/stringtable.c`       | 561 | mostly unchanged; drop primary-write paths that now live in `tp_add_document_terms` |
+| `src/memtable/source.h`            |  18 | rename `tp_memtable_source_create` → `tp_memtable_cache_source_create`; semantics unchanged |
+| `src/memtable/source.c`            | 167 | as above |
+
+What we do **NOT** resurrect:
+
+- `src/replication/replication.h`, `src/replication/rmgr.c`
+  (custom rmgr, 1000+ LOC). The cache is non-WAL'd derived
+  state; replication is handled at the chain layer.
+- `src/index/registry.{c,h}` (484 LOC). The current `state.c`
+  already provides registry-equivalent lifecycle.
+- `src/index/memory.{c,h}` (87 LOC). Memory accounting will be
+  smaller and lives next to the cache.
+
+Net: ~1450 LOC restored. The existing `TpMemtable` scaffold in
+`src/index/state.h` (`string_hash_handle`, `doc_lengths_handle`,
+counters) is already wired through `state.c`; we just fill in the
+handles instead of leaving them invalid.
+
+### Required API change to `tp_memtable_append`
+
+`tp_memtable_append` currently returns `BlockNumber` — the page
+of the just-written record. That's insufficient: the cache hook
+needs (a) the explicit bootstrap signal, (b) the within-page
+offset, (c) the logical "next position" after this record (which
+differs from `(record_blkno, off+1)` for fragment appends because
+the fragment chain ends on a fresh tail page), and (d) the
+`meta.head_blkno` observed at publish time as the generation
+token. This PR changes the return type to
+`TpMemtableAppendResult` (defined in "The apply cursor"
+section). All callers (currently only `tp_add_document_terms`)
+are updated in lockstep; the on-disk format is unchanged.
+
+## Data structures
+
+```c
+/* in src/index/state.h, extending the existing TpMemtable stub */
+typedef struct TpMemtable
+{
+    /* --- inherited v1 fields, currently dead but still allocated --- */
+    dshash_table_handle string_hash_handle;  /* term  → TpStringHashEntry */
+    dshash_table_handle doc_lengths_handle;  /* ctid  → TpDocLengthEntry  */
+    pg_atomic_uint64    total_postings;
+    pg_atomic_uint64    num_terms;
+    pg_atomic_uint64    total_term_len;
+
+    /* --- new in cache design --- */
+
+    /* The single serialization point for cache mutation: every
+     * path that advances the cursor or inserts records into the
+     * cache acquires this EXCL.  Acquired AFTER per-index
+     * LWLock. */
+    LWLock              apply_lock;
+
+    /* Lifetime lock for served TpDataSources: SHARED for the
+     * lifetime of a source (open through close), EXCL only for
+     * full drop / evict.  apply_lock holders may take SHARED
+     * here concurrently with served readers. */
+    LWLock              lock;
+
+    /* Apply cursor (see "The apply cursor" section).  Mutated
+     * only under apply_lock EXCL. */
+    BlockNumber         cursor_gen_head_blkno; /* generation
+                                                * token: equals
+                                                * meta.head_blkno
+                                                * the cache was
+                                                * built against */
+    BlockNumber         cursor_next_blkno;     /* logical next
+                                                * record position */
+    uint16              cursor_next_off;
+    pg_atomic_uint64    cursor_seq;            /* monotonic
+                                                * apply counter,
+                                                * lock-free read */
+
+    /* memory accounting */
+    pg_atomic_uint64    estimated_bytes;       /* cached cache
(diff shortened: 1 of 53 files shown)


docs/memtable_cache.md
```diff
@@ -0,0 +1,923 @@
...
+replaced the v1 shared-memory inverted index with an on-disk
+paged chain of opaque doc records. That change was necessary for
+streaming replication, crash recovery via stock PG replay, and
+single-page WAL-redo compatibility — but it has a real query-time
+cost. Every index scan now opens a `chain_source` that walks
+**every chain page** from `meta.head` to `meta.tail`, decoding
+every doc record, just to materialize the few posting lists the
+query actually needs. With a chain of ~64 pages (the auto-spill
+threshold) that's hundreds of `ReadBuffer` calls per query
+regardless of query selectivity.
```
tjgreen42: Query selectivity is not the right term here.  Maybe keyword frequency is better.

docs/memtable_cache.md
```diff
@@ -0,0 +1,923 @@
...
+  cache** of the same content, organized exactly like the v1
+  memtable: dshash `term → TpPostingList`, dshash `ctid →
+  doc_length`, plus rolling corpus totals.
+- The cache is **derived state**: lazily built on first query,
+  incrementally updated by writers, dropped + rebuilt on spill
+  or invalidation. Crashing or losing it is harmless — the next
+  query rebuilds from the chain.
+- An **apply cursor** tracks the next chain record the cache
+  expects to consume. Three paths can advance the cursor —
+  writer hook, reader catchup, cold lazy build — and all three
```
tjgreen42: Can we simplify and get rid of the writer hook?  Just update the cache lazily on read.

docs/memtable_cache.md
```diff
@@ -0,0 +1,923 @@
...
+  share a single `apply_lock`-serialized helper
+  (`apply_records_up_to`). The cursor IS the commit point;
+  every record is applied to the cache exactly once per
+  generation.
+- The writer hook attaches at the tail of `tp_add_document_
+  terms` while the writer still holds per-index LWLock SHARED.
+  Spill (which needs per-index LWLock EXCL) cannot fire
+  concurrently with cache application, so spill-during-apply is
```
tjgreen42: Would it simplify things to use the cache for spills (and get rid of the code we added for v2 memtable spills, changing it to fill main-memory cache then spill from there)?

docs/memtable_cache.md
```diff
@@ -0,0 +1,923 @@
...
+wrong as soon as fragment append participates. The cursor sidesteps
+this entirely: there is no "before/after" comparison. There is
+only "is the cursor at the current logical tail?", answered by
+asking the chain walker: starting from `next_blkno @ next_off`,
+is there a next record? If not, we're caught up.
+
+### Apply protocol (single shared protocol for all mutators)
+
+Every path that wants to insert a record into the cache —
+whether it's a writer's just-completed append, a reader doing
```
tjgreen42: Prefer to just have read (and spill as a special case of read) mutate the cache

docs/memtable_cache.md
```diff
@@ -0,0 +1,923 @@
...
+                            present + caught up, else chain
+                            source (already the source-of-truth
+                            chooser; no behavioral change for
+                            chain users)
+```
+
+No metapage version bump. No new WAL record types. No on-disk
+format change.
+
+## Testing
```
tjgreen42, line 900: Benchmarking needs to be a part of the testing.  Let's reuse the github workflows we have and adapt to A/B testing of the update workloads with and without the cache.

docs/memtable_cache.md
tjgreen42: Fixed — line 17 now says "regardless of keyword frequency". (commit 364981d1)

docs/memtable_cache.md
tjgreen42: Done — dropped the writer hook entirely. The cache is now updated lazily on read only. This also lets us drop the planned `TpMemtableAppendResult` API change to `tp_memtable_append` — the write path is fully unchanged. The new write-path section just says "unchanged; first reader pays the catchup". (commit 364981d1)

docs/memtable_cache.md
tjgreen42: Done — combined with comment #4 below. `tp_spill` now takes per-index EXCL, calls `apply_to_tail` (or `cold_build` if no cache exists) to bring the cache fully current, then `tp_write_segment_from_cache` flushes it to the segment writer. The v2 `chain_source`-based spill path remains as the fallback when memory cap aborts catchup, but the common path goes through the cache. (commit 364981d1)

docs/memtable_cache.md
tjgreen42: Done — reads (and spill as a special case of read) are now the only cache mutators. The apply protocol has two callers instead of three; `apply_lock` EXCL still serializes them against each other and against eviction. (commit 364981d1)

docs/memtable_cache.md
tjgreen42: Done — added an "A/B benchmarking (via existing workflows)" subsection to Testing. Plan: extend `benchmark.yml` and `benchmark-concurrent-insert.yml` with a `memtable_cache: [off, on]` matrix dimension. Same workload, two legs, metrics tagged per leg. Includes an expected-effect table per workload (insert/sec no-regress, query latency improves, NDCG@10 unchanged). Soft-fail thresholds initially; harden gates after two weeks of nightly baselines. (commit 364981d1)

docs/memtable_cache.md
```diff
@@ -0,0 +1,866 @@
...
+}
+```
+
+The trailing prose about the "cursor caught up check" is now
+obsolete (always catch up first); replace with:
+
+After spill, `tp_cache_clear` deallocates the dshash tables.
+The next query lazy-builds. Because writes don't touch the
+cache, a spill that lands between two queries simply means the
+second query pays a cold-build instead of a catchup walk.
```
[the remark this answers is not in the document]
tjgreen42: Done — addressed in df6119e4. Deleted lines 535-537 (the "The trailing prose about the 'cursor caught up check' is now obsolete; replace with:" marker plus the trailing blank). The paragraph immediately following — "After spill, `tp_cache_clear` deallocates the dshash tables..." — now flows directly after the `tp_spill` pseudocode block.

docs/memtable_cache.md
```diff
@@ -0,0 +1,866 @@
...
+this entirely: there is no "before/after" comparison. There is
+only "is the cursor at the current logical tail?", answered by
+asking the chain walker: starting from `next_blkno @ next_off`,
+is there a next record? If not, we're caught up.
+
+### Apply protocol (single shared protocol for all mutators)
+
+Every path that wants to insert records into the cache —
+reader catchup, cold lazy build, or spill catchup —
+**MUST** follow this protocol:
```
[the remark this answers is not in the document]
tjgreen42: Done — addressed in df6119e4.

Renamed the section "### Apply protocol (catchup)" and scoped the **MUST** clause to "Catchup callers (reader catchup, spill catchup)". The header text now explicitly says: cold lazy build is a **separate** bootstrap path, takes `cache.lock` EXCL rather than SHARED, and is documented in "Read path" under `cold_build`. The two paths share the high-level invariants (apply_lock EXCL serialization, monotonic cursor advance, generation-token check) but the lock modes differ.

Also tightened phase 3 in the implementation plan to call out the two distinct paths and their different `cache.lock` modes.

docs/memtable_cache.md
```diff
@@ -0,0 +1,866 @@
...
+  threshold.
+- **`get_postings(term)`** on the returned source does
+  `tp_string_table_lookup` under dshash bucket lock, acquires
+  `TpPostingList.lock` SHARED, copies entries into heap-
+  allocated `TpPostingData`, releases. Identical to v1.
+- **Hold `cache.lock` SHARED across the entire scan**. This
+  excludes drop/evict for the source's lifetime — necessary
+  because the source's served `TpPostingData` references DSA
+  pointers that drop would invalidate. Per-query LIMIT (default
+  1000) bounds the lifetime.
```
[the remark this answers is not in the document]
tjgreen42: Done — addressed in df6119e4.

Rewrote the two bullets at the bottom of the Read path "Notes" section. `get_postings` is now described as returning a fully-copied, self-contained `TpPostingData` of POD values (no surviving DSA references), and the lifetime-spanning `cache.lock` SHARED is now justified by the source's cached `dshash_table` attachments (`string_table`, `doclength_table`) — `get_postings` does a fresh dshash lookup per query term, and BM25 scoring calls `get_doc_length` per posting hit, so a concurrent `tp_cache_clear` that reset the handles to INVALID and freed the underlying tables would dangle those attachments. Wording follows your suggested phrasing.

docs/memtable_cache.md
```diff
@@ -0,0 +1,866 @@
...
+
+     /* re-check after lock acquire: another backend may have
+      * evicted the same target. */
+     if cache still present and estimated_bytes >= threshold:
+         tp_cache_clear(target)
+         atomic_sub(global estimated_total_bytes,
+                    target.estimated_bytes)
+
+     release in reverse order
+   ```
```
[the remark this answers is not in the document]
tjgreen42: Done — addressed in df6119e4.

Adopted recommendation (a)+(c): `evict_largest` now syntactically excludes the caller's own index from the argmax (`argmax_{idx ≠ caller_index}`), and a new `global_eviction_mutex` is acquired before any target's per-index lock to serialize evictions across backends. The lock-order section adds `global_eviction_mutex` above any **other**-index per-index LWLock, with the constraint `target ≠ caller` enforced syntactically.

Also: moved the global-cap check out of the mid-walk loop entirely. The check now runs only at apply-protocol entry, before the caller acquires `cache.apply_lock`/`cache.lock` — keeps the caller's own cache locks out of the eviction acquisition chain. The mid-walk per-record check that remains is purely a per-index soft-cap check that returns BUDGET_EXCEEDED without triggering eviction.

Added an explicit subsection for the case where the caller's own index is the largest: `evict_largest` returns NOTHING_TO_EVICT, caller returns BUDGET_EXCEEDED, read falls back to `chain_source`. A different backend on a later query will pick this index up.

Issue #404, opened 2026-06-01
Partial BM25 reads can fail with invalid segment header during concurrent matching writes and spill/merge


Lucasbjorklund31:
## Bug Description

`pg_textsearch 1.3.0` can fail reads from a **selective partial BM25 index** when matching writes run concurrently with `bm25_spill_index()` or `bm25_force_merge()`.

## Environment

- **PostgreSQL version**: 18.0
- **pg_textsearch version**: 1.3.0
- **Operating system**: Debian 13 / trixie container, via `ghcr.io/cloudnative-pg/postgresql:18.0-system-trixie`

## Steps to Reproduce

1. Create a table with a full BM25 index and a selective partial BM25 index.
2. Insert initial rows and spill/merge both indexes.
3. Concurrently run:
   - inserts/updates matching the partial-index predicate,
   - `bm25_spill_index()` / `bm25_force_merge()`,
   - reads from the partial BM25 index.
4. The partial-index reader can fail with an invalid segment header.

```bash
#!/usr/bin/env bash
set -u

PSQL='psql -h 127.0.0.1 -U postgres -d postgres -qAt -v ON_ERROR_STOP=1'

$PSQL <<'SQL'
SET client_min_messages=warning;
CREATE EXTENSION IF NOT EXISTS pg_textsearch;

DROP TABLE IF EXISTS docs;
CREATE TABLE docs (id bigserial PRIMARY KEY, content text NOT NULL, lang text NOT NULL);

CREATE INDEX docs_bm25 ON docs USING bm25(content) WITH (text_config='simple');
CREATE INDEX docs_de_bm25 ON docs USING bm25(content) WITH (text_config='german') WHERE lang = 'de';

INSERT INTO docs (content, lang)
SELECT repeat('postgres bm25 search ', 200) || gs,
       CASE WHEN gs % 2 = 0 THEN 'de' ELSE 'en' END
FROM generate_series(1, 4000) gs;

SELECT bm25_spill_index('docs_bm25');
SELECT bm25_spill_index('docs_de_bm25');
SELECT bm25_force_merge('docs_bm25');
SELECT bm25_force_merge('docs_de_bm25');
SQL

writer() {
  for i in $(seq 1 100); do
    $PSQL -c "INSERT INTO docs (content, lang)
      SELECT 'writer postgres bm25 ' || gs, CASE WHEN gs % 2 = 0 THEN 'de' ELSE 'en' END
      FROM generate_series(1, 20) gs" || exit 10
[...]

PR #412, opened 2026-06-15
Fix VACUUM-vs-merge race: read segment metapage under the per-index lock (#411)

### LINKED ISSUE #411: VACUUM/autovacuum of a bm25 index races with concurrent spill/merge: invalid segment header + out-of-bounds alive bitset (Assert(doc_id < num_docs))
## Bug Description

`VACUUM` (including **autovacuum**) of a table with a `bm25` index races with
concurrent segment **spill/merge**, because the index bulk-delete path walks
segments **by block number without holding the per-index lock**, unlike the
read path that was hardened in #404 (`dcbb806`).

This has two manifestations, both from the same root cause:

1. **Frequent, easy to reproduce:** `VACUUM` fails with
   `ERROR: invalid segment header at block N` (`while cleaning up index ...`)
   when a concurrent spill/merge recycles a block the vacuum is reading.
2. **Severe escalation:** when a recycled block still passes the segment-header
   magic check but the segment was **shrunk by a merge**, a stale per-segment
   `doc_id` collected earlier in the same `VACUUM` is now `>= bitset->num_docs`,
   and `tp_alive_bitset_mark_dead()` writes **out of bounds**:
   - cassert build → `TRAP: failed Assert("doc_id < bitset->num_docs")`
     (`src/segment/alive_bitset.c:101`)
   - release build → out-of-bounds heap write at
     `src/segment/alive_bitset.c:109` (`bitset->bits[doc_id >> 3] &= ~mask`)
     → heap corruption / **SIGSEGV**.

This is **distinct from #404**: that fix only hardened the standalone-scoring /
read path (`src/types/query.c`, `src/debug/dump.c`). The `VACUUM` bulk-delete
path (`src/access/vacuum.c`) was not touched and is still unprotected on `main`
(`876497f`): `tp_alive_bitset_mark_dead`'s assert and the
`tp_vacuum_identify_affected` → `tp_vacuum_mark_dead` open-by-block-number flow
are unchanged.

## Environment

- **PostgreSQL version**: 17.9 (stock, from postgresql.org; `--enable-debug
  --enable-cassert`)
- **pg_textsearch version**: 1.3.0 (code path identical on `main` @ `876497f`)
- **Operating system**: Azure Linux 3 (x86_64), gcc 13.2.0

## Steps to Reproduce

Build PostgreSQL with `--enable-cassert`, `shared_preload_libraries =
'pg_textsearch'`, aggressive autovacuum (`autovacuum_naptime = 1s`, per-table
`autovacuum_vacuum_threshold = 200`, `autovacuum_vacuum_scale_factor = 0`).

```sql
CREATE EXTENSION pg_textsearch;
CREATE TABLE cc (id bigserial PRIMARY KEY, body text);
CREATE INDEX cc_idx ON cc USING bm25(body) WITH (text_config='english');
ALTER TABLE cc SET (autovacuum_vacuum_scale_factor=0, autovacuum_vacuum_threshold=200);
INSERT INTO cc(body) SELECT 'seed '||md5(g::text) FROM generate_series(1,2000) g;
```

Then drive concurrent load for a few minutes (e.g. with `xargs -P` or many
`psql` sessions):

- **6 writers** (insert + immediate spill → many segments → merges):
  ```sql
  SET pg_textsearch.memtable_pages_threshold = 4;
  INSERT INTO cc(body) SELECT 'w '||md5(random()::text) FROM generate_series(1,200) g;
  SELECT bm25_spill_index('cc_idx');
  ```
- **1 deleter** (monotonic, oldest-first → creates dead docs, no row-lock
  deadlocks): `DELETE FROM cc WHERE id IN (SELECT id FROM cc ORDER BY id ASC LIMIT 120);`
- **2 readers**: `SELECT count(*) FROM (SELECT 1 FROM cc ORDER BY body <@> to_bm25query('w','cc_idx') LIMIT 20) s;`
- autovacuum drives the `VACUUM` bulk-delete.

## Expected Behavior

`VACUUM`/autovacuum of a `bm25` index is safe under concurrent spill/merge: no
`invalid segment header` errors and no out-of-bounds access.

## Actual Behavior

Within ~1 minute, hundreds of:

```
ERROR:  invalid segment header at block N
DETAIL:  magic=0x5450544D, expected 0x54505347
CONTEXT:  while cleaning up index "cc_idx" of relation "public.cc"
STATEMENT:  VACUUM cc;
```

(`0x54505347` = `TP_SEGMENT_MAGIC` "TPSG"; observed `0x5450544D` =
`TP_MEMTABLE_PAGE_MAGIC` "TPTM" → the block was repurposed by a concurrent
spill/merge mid-vacuum.)

On a cassert build, the escalation crashes an autovacuum worker:

```
TRAP: failed Assert("doc_id < bitset->num_docs"),
      File: "src/segment/alive_bitset.c", Line: 101
LOG:  server process (PID ...) was terminated by signal 6: Aborted
DETAIL:  Failed process was running: autovacuum: VACUUM public.cc
```

Backtrace (cassert build; core had `bitset->num_docs == 2` while a stale
`doc_id >= 2` was being marked):

```
#3 ExceptionalCondition (... "doc_id < bitset->num_docs" ... assert.c:66)
#4 tp_alive_bitset_mark_dead (bitset=..., doc_id=<optimized out>) at src/segment/alive_bitset.c:101
#5 tp_vacuum_mark_dead (...) at src/access/vacuum.c:777
#6 tp_bulkdelete (...) at src/access/vacuum.c:919
#7 index_bulk_delete (...) at indexam.c:758
   ... -> heap_vacuum_rel -> autovacuum_do_vac_analyze (autovacuum worker)
```

## Root Cause

`tp_bulkdelete` (`src/access/vacuum.c`) is a two-phase operation:

- **Phase 2** `tp_vacuum_identify_affected()` walks segments and collects, per
  segment, the list of dead `doc_id`s.
- **Phase 3** `tp_vacuum_mark_dead()` (`vacuum.c:751`) re-opens each segment **by
  block number** (`tp_segment_open_ex(index, root_block, ...)`),
  `tp_alive_bitset_load()`s it, and marks each collected `doc_id`.

Neither phase holds the per-index lock across the operation, so a concurrent
`bm25_spill_index()` / `bm25_force_merge()` / auto-spill can free and recycle a
segment's blocks (or replace the segment with a smaller merged one) between
Phase 2 and Phase 3. The reopened block is then either a different page type
(→ `invalid segment header`) or a valid-but-smaller segment (→ stale `doc_id`
out of range → the assert/OOB).

## Suggested Fix

Mirror the #404 (`dcbb806`) interlock in the `VACUUM` path: hold the per-index
lock `LW_SHARED` and re-read the metapage / per-level segment roots **under the
lock** across the identify+mark of `tp_bulkdelete`, so a concurrent spill/merge
(which takes `LW_EXCLUSIVE`) cannot recycle the blocks mid-vacuum. A defensive
bounds check in `tp_alive_bitset_mark_dead` (skip `doc_id >= num_docs` instead of
asserting / writing OOB) is also warranted as defense-in-depth for release
builds.

## Additional Context

- Note for CI: `make installcheck` runs the regression tests **serially**, and
  the dedicated sanitizer workflow also runs `installcheck` serially, so neither
  can catch this **concurrency** bug. A concurrent stress test (writers +
  autovacuum on one shared `bm25` index, like
  `test/scripts/partial_concurrent_read.sh` from #404 but exercising VACUUM)
  would guard it.

tjgreen42:
## Summary

Fixes #411 — a TOCTOU race between `VACUUM`/autovacuum and concurrent segment spill/merge that produced `invalid segment header` errors and, in the worst case, an out-of-bounds write in the alive bitset (`TRAP: failed Assert("doc_id < bitset->num_docs")` under cassert, or a SIGSEGV in release builds).

The race affects **both** VACUUM paths in `src/access/vacuum.c`. Both read the metapage `level_heads` snapshot and then open segments **by block number**, while a concurrent spill/merge/compaction (`tp_do_spill` → `tp_maybe_compact_level`, or `bm25_force_merge`, all under `LW_EXCLUSIVE`) frees and recycles those blocks via the FSM:

- **`tp_bulkdelete`** (`ambulkdelete`, `CONTEXT: while vacuuming index`) held **no** per-index lock across Phase 2 (`tp_vacuum_identify_affected`) → Phase 3 (`tp_vacuum_mark_dead`). A segment shrunk by a merge between the phases makes a previously-collected `doc_id` exceed the reloaded bitset's `num_docs`, driving the out-of-bounds write in `tp_alive_bitset_mark_dead`.
- **`tp_vacuumcleanup`** (`amvacuumcleanup`, `CONTEXT: while cleaning up index`) *did* take the per-index lock `LW_SHARED`, but **acquired it after reading the metapage**, so the snapshot walked by `tp_count_live_docs` was already stale. Under autovacuum this is the dominant manifestation of `invalid segment header`.

## Changes

- **`tp_bulkdelete`**: acquire the per-index lock `LW_SHARED` after Phase 1's spill (which itself takes `LW_EXCLUSIVE`, so this avoids a shared→exclusive upgrade) and hold it across identify + mark/replace; release on every exit path.
- **`tp_vacuumcleanup`**: acquire `LW_SHARED` **before** reading the metapage so the `level_heads` snapshot is taken under the lock.
- **`tp_alive_bitset_mark_dead`**: skip `doc_id >= num_docs` instead of writing out of bounds — defense in depth for non-assert builds (the assert is preserved as a canary).

Inserts and scans already hold this lock `LW_SHARED`, so they are unaffected; only the exclusive segment-recyclers wait. This matches the existing convention already used by `tp_vacuumcleanup`'s reclaim walk and the #404 read-path fix.

## Test

Adds `test/scripts/vacuum_concurrent_merge.sh` (wired into `make test-concurrency`): concurrent writers (spilling many small segments) + `bm25_force_merge` loop + monotonic deleter + `VACUUM` loop on one shared `bm25` index, with aggressive autovacuum. It checks the server and client logs for `invalid segment header` and crash signatures. A serial `installcheck` cannot catch this concurrency bug, so a dedicated stress script is needed.

## Verification (stock PostgreSQL 17.9, `--enable-cassert`)

A/B with a concurrent harness (6 writers + force-merge loop + monotonic deleter + 2 readers, aggressive autovacuum):

| build | duration | `invalid segment header` (VACUUM ctx) | crash |
|---|---|---|---|
| unmodified `main` | 4 min | **18** | — |
| `tp_bulkdelete` fix only | 10 min | 24 (all cleanup ctx) | — |
| **both paths fixed (this PR)** | 10 min | **0** | none |

- `vacuum_concurrent_merge.sh`: **fails within ~30 s on `main`**, **passes** with this change.
- `make installcheck`: all 70 tests pass.
- `make format-check`: clean on the changed files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,13 @@ jobs:
           echo "✗ Concurrency tests failed"
           exit 1
         fi
+        echo "Running VACUUM concurrency tests..."
+        if (cd test/scripts && ./vacuum_concurrent_merge.sh); then
+          echo "✓ VACUUM concurrency tests passed"
+        else
+          echo "✗ VACUUM concurrency tests failed"
+          exit 1
+        fi
         echo "Running segment tests..."
         if (cd test/scripts && ./segment.sh); then
           echo "✓ Segment tests passed"
--- a/.github/workflows/sanitizer-build-and-test.yml
+++ b/.github/workflows/sanitizer-build-and-test.yml
@@ -185,6 +185,14 @@ jobs:
           cd test/scripts && ./concurrency.sh
         "
 
+        echo "Running VACUUM concurrency tests under sanitizer..."
+        timeout 1200s bash -c "
+          export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
+          export PGHOST=$PWD/tmp_sanitizer_test
+          export PGPORT=55433
+          cd test/scripts && ./vacuum_concurrent_merge.sh
+        "
+
         echo "Running crash recovery tests under sanitizer..."
         timeout 600s bash -c "
           export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,7 @@ test-concurrency:
 	@echo "Running concurrency tests..."
 	@cd test/scripts && ./concurrency.sh
 	@cd test/scripts && ./partial_concurrent_read.sh
+	@cd test/scripts && ./vacuum_concurrent_merge.sh
 
 test-recovery:
 	@echo "Running crash recovery tests..."
--- a/src/access/vacuum.c
+++ b/src/access/vacuum.c
@@ -851,7 +851,28 @@ tp_bulkdelete(
 	if (index_state != NULL)
 		tp_spill_memtable_if_needed(info->index, index_state, 1);
 
-	/* Re-read metapage after spill */
+	/*
+	 * Hold the per-index LWLock in shared mode across Phase 2 (identify)
+	 * and Phase 3 (mark / replace).  A concurrent spill / merge /
+	 * compaction takes LW_EXCLUSIVE to mutate level_heads and free
+	 * segment pages via the FSM, so without this lock a segment we
+	 * identify in Phase 2 can be shrunk or have its blocks recycled
+	 * before Phase 3 reopens it by block number -- yielding an
+	 * "invalid segment header" error or, when a shrunk segment still
+	 * passes the header magic check, an out-of-bounds alive-bitset
+	 * write from a now-stale doc_id (tp_alive_bitset_mark_dead).  See
+	 * tp_vacuumcleanup, which holds the same lock for the same reason.
+	 *
+	 * Acquire after Phase 1's spill (which takes LW_EXCLUSIVE itself) to
+	 * avoid a shared->exclusive upgrade, and before re-reading the
+	 * metapage so the level_heads snapshot we walk stays stable.  Inserts
+	 * and scans also hold this lock LW_SHARED, so they are unaffected;
+	 * only the exclusive block-recyclers wait.
+	 */
+	if (index_state != NULL)
+		tp_acquire_index_lock(index_state, LW_SHARED);
+
+	/* Re-read metapage after spill (now under the shared lock) */
 	pfree(metap);
 	metap = tp_get_metapage(info->index);
 	if (!metap)
@@ -859,6 +880,7 @@ tp_bulkdelete(
 		stats->num_pages		= 1;
 		stats->num_index_tuples = 0;
 		stats->tuples_removed	= 0;
+		tp_release_index_lock(index_state);
 		return stats;
 	}
 
@@ -880,6 +902,7 @@ tp_bulkdelete(
 		stats->pages_deleted	= 0;
 		pfree(metap);
 		pfree(segments);
+		tp_release_index_lock(index_state);
 		return stats;
 	}
 
@@ -1002,6 +1025,9 @@ tp_bulkdelete(
 				info->index, docs_shrinkage, tokens_shrinkage);
 	}
 
+	/* Identify + mark complete; drop the shared lock. */
+	tp_release_index_lock(index_state);
+
 	/*
 	 * tp_vacuumcleanup will set num_index_tuples to the actual live
 	 * count; only tuples_removed needs to carry through from here.
@@ -1050,7 +1076,20 @@ tp_vacuumcleanup(IndexVacuumInfo *info, IndexBulkDeleteResult *stats)
 		tp_spill_memtable_if_needed(
 				info->index, index_state, TP_MIN_SPILL_PAGES);
 
-	/* Get current index statistics from metapage */
+	/*
+	 * Acquire the per-index LWLock in shared mode *before* reading the
+	 * metapage so the level_heads snapshot we walk in tp_count_live_docs
+	 * stays valid: concurrent spills / compactions take LW_EXCLUSIVE to
+	 * mutate level_heads and free segment pages via the FSM, so a snapshot
+	 * read outside the lock could leave us opening a block a merge has
+	 * already recycled ("invalid segment header").  Acquire after the
+	 * spill above (which takes LW_EXCLUSIVE itself) to avoid a
+	 * shared->exclusive upgrade.
+	 */
+	if (index_state != NULL)
+		tp_acquire_index_lock(index_state, LW_SHARED);
+
+	/* Get current index statistics from metapage (under the shared lock) */
 	metap = tp_get_metapage(info->index);
 	if (metap)
 	{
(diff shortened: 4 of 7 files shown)


tjgreen42:
### Code review pass (self-review) + follow-ups addressed

Ran a thorough concurrency/correctness review of this change. Summary and the resulting updates (pushed in the follow-up commit):

**Core fix verified sound.** Traced the locking model end-to-end: no lock-order inversion (per-index LWLock is always taken before buffer locks), no lock leak on the `ereport(ERROR)` path (`AbortTransaction` → `LWLockReleaseAll` then `XACT_EVENT_ABORT` → `tp_release_all_index_locks`, with the `LWLockHeldByMe` re-check reconciling `lock_held`), and no hidden `LW_SHARED`→`LW_EXCLUSIVE` upgrade inside the protected region (the only `EXCLUSIVE` acquire is Phase 1's spill, which runs and releases before the `SHARED` region). The VACUUM-vs-merge race (#411) is genuinely closed for both `tp_bulkdelete` and `tp_vacuumcleanup`.

**Changes made in response to review:**
1. **CI now actually runs the new test.** CI hand-lists the concurrency scripts instead of `make test-concurrency`, so the regression guard would not have gated PRs. Wired `vacuum_concurrent_merge.sh` into `ci.yml` and the **sanitizer** job (the latter deterministically catches the OOB write).
2. **Corrected an over-claiming comment.** The Phase-2/3 comment said concurrent scans are "unaffected." That's not accurate: VACUUM's own Phase-3 page reclamation (`tp_vacuum_replace_segment` → `RecordFreeIndexPage`, on all-dead drops / pre-V5 rebuilds) also runs under `LW_SHARED`, which does not exclude concurrent `LW_SHARED` scans. This is **pre-existing** (pre-patch Phase 3 held no lock at all) and out of #411's scope — filed as #413 and referenced from the comment.
3. Updated the `tp_alive_bitset_mark_dead` header-prototype doc to match the implementation, and guarded the three new `tp_release_index_lock()` calls for consistency with `tp_vacuumcleanup`.

**Verification:** `make installcheck` 70/70 on a clean cluster; `vacuum_concurrent_merge.sh` fails (`invalid segment header` within ~30s) on `main` and passes on this branch; `clang-format` clean on changed lines.

**Known residual (separate, pre-existing):** the VACUUM-vs-scan Phase-3 reclamation race (#413) and a debug-only unprotected `pageviz` walk — both out of scope here.

PR #425, opened 2026-07-06
docs: recommend periodic REINDEX for fragmented indexes

mory91:
## Summary

Adds a short subsection to the README's Performance Tuning section noting
that update-heavy workloads can fragment an index over time, and that a
periodic `REINDEX` restores a contiguous page layout.

That is the entire change: **15 lines of README, no code.**

## How this PR got here

It started as an implementation of segment defragmentation — a two-phase
merge (segment → BufFile → contiguous page range) so that `bm25_force_merge`
would produce physically sequential pages, behind a
`pg_textsearch.defrag_on_merge` GUC. That code has been dropped. The
benchmarking done to justify it is what led here, and it's worth recording,
because it answers two separate questions:

**Does fragmentation hurt? Enormously — when it exists.**
An experimental `debug_randomize_pages` GUC pre-extended the relation and
served page allocations in shuffled order, giving a worst-case layout. Two
otherwise identical bulk builds over MS MARCO v2 (138M passages, 691
multi-token queries, cold cache) differed only in physical page order:

| | Normal | Randomized |
|---|---|---|
| Weighted avg latency | 63.8 ms | 594.4 ms (**9.3x**) |
| p99, 8+ token queries | 558 ms | 12,513 ms (**22x**) |

Multi-token queries degraded 5-12x. ([full results](https://github.com/timescale/pg_textsearch/pull/425#issuecomment-5331169374))

**Does it happen in practice? Not from normal operation.**
An incremental build of the same dataset (138M rows in 1M-row batches, 135
forced spills, 12 segments across L0-L5) followed by a cross-level force
merge moved the weighted average from 63.98 ms to 63.66 ms — 0.5%, inside
the noise. When no free pages are available the FSM extends the relation
sequentially, so the large segments were already laid out contiguously and
defragmentation had nothing to fix. An earlier run on the standard build
found force merge was a literal no-op, because auto-compaction had already
left one segment per level and `tp_force_merge_all` needs two at a level to
do anything. ([details](https://github.com/timescale/pg_textsearch/pull/425#issuecomment-5291722505))

So the defrag path worked correctly but addressed a state the allocator
doesn't currently produce, at the cost of a second write path, ~59% added
force-merge time, and index bloat — force merge grew the relation 28 GB →
45 GB, since displaced pages are tombstoned for deferred reclaim rather
than freed in place.

Given that, and that `REINDEX` already recovers a clean layout for the
pathological cases, the conclusion was that the complexity isn't warranted
yet. Hence: document the workaround, drop the code.

## Follow-ups this surfaced

- Contiguous layout is better baked into the compaction write path than
  bolted onto force merge. Worth revisiting under background
  compaction / index health maintenance.
- `bm25_force_merge` only merges *within* a level, so it cannot consolidate
  to a single segment once auto-compaction has left one per level. Note that
  the README's Force-merging section still claims it "rewrites all segments
  into a single segment", which is inaccurate; worth a separate fix.
- The `debug_randomize_pages` GUC and the benchmark scripts are useful for
  validating any future layout work, and are preserved in the PR history.

Thanks @mory91 for the implementation and for the several rounds of
benchmarking that produced this result — the negative result plus the
worst-case bound is more useful than the feature would have been.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,21 @@ This is analogous to Lucene's `forceMerge(1)`. It rewrites all segments into
 a single segment and reclaims the freed pages. Best used after large batch
 inserts, not during ongoing write traffic.
 
+#### Index fragmentation on update-heavy workloads
+
+Sustained INSERT/DELETE/UPDATE traffic can cause index pages to become
+physically scattered across the relation file, reducing sequential-read
+efficiency. If cold-cache query latency degrades over time, a periodic
+`REINDEX` rebuilds the index with contiguous page layout:
+
+```sql
+REINDEX INDEX docs_idx;
+```
+
+This is the same as `DROP INDEX` + `CREATE INDEX` but keeps the index
+name. Schedule it during low-traffic windows since it takes an
+exclusive lock.
+
 #### Use LIMIT with ORDER BY
 
 Top-k queries (`ORDER BY ... LIMIT n`) enable Block-Max WAND optimization,


tjgreen42:
@mory91 thanks much for the PR, and apologies for the slow response.  I'm on vacation for a few weeks.  Can you please include an assessment of the performance impact of this PR?  I noticed you added some benchmarks, but I'd like to see some numbers to see whether the measurements align with intuition about defragmentation.

mory91:
@tjgreen42 thanks for the check. I added Performance Assessment section.

tjgreen42:
@mory91 The query numbers speedups look promising, thanks for measuring that.  One more question, what is the impact on force merge performance?

mory91:
@tjgreen42 I've updated the report with the force merge benchmark results.

tjgreen42:
@mory91 Hmm, cold query perf seems to benefit, but warm query perf (common case) does not seem to be improved. At the same time the scale of the experiment is also too small to tell much.  Sorry to be a pain about this but since there are real costs in terms of merge time (plus code complexity), we need more evidence before greenlighting this one.  Can you repeat the experiment with MS-MARCO v2?  I'd also like to see perf numbers for multi-token queries.  So I'd suggest you run a variant of MS-MARCO v2 that introduces fragmentation (e.g. by building the index incrementally in chunks rather than in bulk), take query timings for the standard MS-MARCO v2 mix, then force-merge and take another round of query timings.

mory91:
## MS-MARCO v2 Benchmark Results (138M passages)

Ran the defrag benchmark on an Ubuntu VM with PostgreSQL 17.11, `shared_buffers=128MB`, against the full MS-MARCO v2 dataset (138M passages). The index was built incrementally in 1M-row batches with forced spills to create fragmentation across 3 levels (L3, L4, L6).

### Setup

- **Dataset**: MS-MARCO v2 (138,364,158 passages)
- **Index**: 44 GB on disk, 3 segments (L3: 305K docs, L4: 8M docs, L6: 130M docs)
- **Queries**: 691 multi-token queries across 8 token buckets
- **Cold cache**: PostgreSQL restarted + OS page cache purged between runs

### Result: force merge was a no-op

Force merge completed in **663 ms** and produced **no change** — pre- and post-merge index summaries are identical. The reason: each level had exactly 1 segment, and `tp_force_merge_all` requires ≥2 segments at a level to merge. The incremental build with spills triggered auto-compaction (`segments_per_level`), which had already merged segments up through the levels before force merge ran.

### Query latency (warm cache, scattered vs "contiguous")

Since nothing was actually defragmented, these numbers are just noise (same index, two measurements):

| Bucket | n | Scattered p50 | Contiguous p50 | Δ% |
|--------|---|---------------|----------------|-----|
| 1 token | 100 | 5.25 ms | 5.32 ms | -1.2% |
| 2 tokens | 100 | 10.29 ms | 9.99 ms | +2.8% |
| 3 tokens | 100 | 24.41 ms | 22.97 ms | +5.9% |
| 4 tokens | 100 | 51.93 ms | 50.71 ms | +2.4% |
| 5 tokens | 100 | 79.53 ms | 79.12 ms | +0.5% |
| 6 tokens | 100 | 119.47 ms | 117.19 ms | +1.9% |
| 7 tokens | 53 | 193.84 ms | 193.13 ms | +0.4% |
| 8+ tokens | 38 | 214.75 ms | 211.36 ms | +1.6% |

Weighted avg: 57.05 ms → 56.07 ms (within noise).

### What we tried: merging across empty levels

We also experimented with changing `tp_force_merge_all` to `continue` instead of `break` when a level has <2 segments, so it scans all levels rather than stopping at the first empty one. This allowed force merge to actually consolidate segments, but results were worse — the merged single segment produced longer posting lists that reduced BMW's ability to skip blocks, and the tombstone system caused the index to bloat from 31 GB to 47 GB (old pages deferred-freed, trapped mid-file). That change has been reverted.

### Core issue

The defragmentation primitive itself (BufFile → contiguous pages) works correctly, but we haven't found a benchmark scenario where it produces a measurable improvement:

1. **Auto-compaction prevents fragmentation**: By the time force merge runs, segments have already been compacted into 1-per-level — nothing to merge.
2. **Single-segment defrag not supported**: Force merge can't rewrite a lone segment contiguously without a merge partner.
3. **Tombstone bloat**: When force merge does consolidate, the deferred page-freeing system prevents the file from shrinking, offsetting layout gains.

What do you think? Should we explore single-segment defragmentation (rewrite in place without merging), or is this feature not worth the complexity given that auto-compaction already keeps things fairly tidy?

tjgreen42:
@mory91 Thanks for running the benchmark and generally being patient with my demands on this PR :)  Indeed, the numbers you've gathered so far are suggesting this one might not be worth it.  But if you have energy to push a little more:

1. I didn't realize that force-merge does not merge across levels.  We should definitely expose some way to do this, so that you can produce a single segment, same as bulk indexing produces.  I will add this to my to-do list for the background compaction work item.  But in the meantime, you might just hack this up to test the effect of your defragmentation code better.

2. It's also possible that your experiment is not producing enough fragmentation to matter for queries.  Try turning the fragmentation up to 11 by writing some code just for the experiment that e.g. pre-allocates some suitably huge number of pages, then serves page-allocation requests during index building by returning pages in a random order.  This would give an upper-bound on the effects of fragmentation.

One last thought, it occurred to me that to defragment the index, you can also just drop/create it.  This takes longer than the alternative you have implemented due to re-tokenization work etc, but it is a serviceable workaround.

mory91:
## Fragmentation experiments: upper bound + real-world incremental build

We ran two experiments on the full MS-MARCO v2 dataset (138M passages) to answer: does page fragmentation matter, and does it occur in practice?

**Setup**: c5d.2xlarge (8 vCPU, 16 GB RAM, NVMe SSD), PostgreSQL 17.11, `shared_buffers = 4GB`, cold-cache between measurements (PG restart + OS page cache purge), 691 multi-token MS-MARCO queries across 8 token buckets.

### Experiment 2: Upper bound (worst-case fragmentation)

We added a `debug_randomize_pages` GUC that pre-extends the relation and serves page allocations in Fisher-Yates shuffled order. Two identical bulk builds on 138M passages — one normal (contiguous), one with fully randomized page layout. Same segment structure, same posting lists, only physical page order differs.

| Tokens | n | Normal p50 | Random p50 | Slowdown | Normal avg | Random avg | Slowdown |
|--------|---|-----------|-----------|----------|-----------|-----------|----------|
| 1 | 100 | 7.7 ms | 8.4 ms | 1.1x | 7.9 ms | 9.8 ms | 1.2x |
| 2 | 100 | 12.7 ms | 25.5 ms | 2.0x | 16.9 ms | 98.5 ms | 5.8x |
| 3 | 100 | 26.9 ms | 128.3 ms | 4.8x | 33.5 ms | 325.9 ms | 9.7x |
| 4 | 100 | 56.3 ms | 259.4 ms | 4.6x | 67.2 ms | 600.8 ms | 8.9x |
| 5 | 100 | 89.8 ms | 409.7 ms | 4.6x | 104.8 ms | 990.7 ms | 9.5x |
| 6 | 100 | 132.1 ms | 916.0 ms | 6.9x | 161.3 ms | 1399.9 ms | 8.7x |
| 7 | 53 | 200.5 ms | 2071.6 ms | 10.3x | 222.7 ms | 2389.5 ms | 10.7x |
| 8+ | 38 | 242.6 ms | 2381.4 ms | 9.8x | 246.7 ms | 3041.6 ms | 12.3x |

**Weighted avg: 63.8 ms vs 594.4 ms — 9.3x slower with scattered pages.** Tail latencies are even worse (p99 for 8-token: 558 ms vs 12,513 ms).

Index sizes: 17 GB (normal) vs 18 GB (randomized) — same 2.2M live pages.

**Conclusion**: When fragmentation exists, the impact is massive — 5-12x for multi-token queries on a dataset larger than RAM.

### Experiment 1: Real-world incremental build → force merge

We also implemented a cross-level force merge that collects segments from ALL levels and merges them in a single N-way pass (previously `bm25_force_merge` only merged within each level, which was a no-op when auto-compaction left 1 segment per level).

Incremental build: 138M rows in 1M-row batches with forced spills (135 spills), producing 12 segments across L0-L5 (28 GB on disk). Force merge consolidated into 1 L6 segment (17.3 GB live data, 45 GB on disk due to tombstone-parked displaced pages). Merge time: 42 minutes.

| Tokens | Scattered p50 | Contiguous p50 | Change |
|--------|--------------|----------------|--------|
| 1 | 7.21 ms | 7.60 ms | -5.4% |
| 2 | 13.01 ms | 12.57 ms | +3.4% |
| 3 | 26.96 ms | 26.89 ms | +0.3% |
| 4 | 56.24 ms | 56.54 ms | -0.5% |
| 5 | 89.75 ms | 89.07 ms | +0.8% |
| 6 | 131.90 ms | 132.40 ms | -0.4% |
| 7 | 202.96 ms | 200.35 ms | +1.3% |
| 8+ | 242.63 ms | 241.68 ms | +0.4% |

**Weighted avg: 63.98 ms → 63.66 ms (0.5% — within noise).**

**Conclusion**: The incremental build with spills + auto-compaction doesn't produce meaningful page fragmentation. The FSM extends the relation sequentially when no free pages are available, so the L5 segments (~8 GB each) were already laid out mostly contiguously. Force merge + defrag had nothing to fix.

### Summary

| | Fragmentation impact | Occurs in practice? |
|---|---|---|
| **Worst-case** (Exp 2) | 9.3x slowdown | Only with deliberately randomized allocation |
| **Incremental build** (Exp 1) | 0.5% (noise) | FSM/extend keeps pages sequential |

Page fragmentation can be devastating when it exists, but the current allocation strategy doesn't produce it during normal operations. The defrag code works correctly but solves a problem that doesn't currently arise. The main scenario where it could matter is heavy delete/vacuum cycles that scatter free pages throughout the file, which are then reused for new segments — but we haven't measured that case.

The cross-level force merge itself (independent of defragmentation) is potentially useful as a way to produce a single segment like bulk indexing, though the 28→45 GB tombstone bloat and 42-minute merge time are costs to consider.

tjgreen42:
@mory91 THanks for your continued diligence on this task.  I think the bottom line is: (1) fragmentation hurts A LOT when maximized, as shown in your experiments; (2) for now, we have a workaround in the form of REINDEX that performs well enough.  WDYT?

Might be good to revisit as part of the background compaction/health maintenance task.

mory91:
REINDEX (drop/create) is a serviceable workaround for pathological cases, without the added complexity of a two-phase BufFile write path.

Happy to revisit this as part of background compaction/health maintenance, where contiguous layout can be baked into the compaction write path naturally. The experimental GUC (debug_randomize_pages) and benchmark scripts could be useful for validating that work.

Thanks for the thorough review and patience with the benchmarking iterations!

tjgreen42:
Okay, I think on balance the extra complexity/LOC is not worth it for now.  But we learned good lessons from your efforts here, thanks again!  Maybe the last thing we could is add a piece of guidance to the README advising folks that update-heavy workloads can yield fragmented indexes over time, so periodic REINDEX can help performance.

mory91:
@tjgreen42 thanks alot for your guidance. I added the change to the README.

tjgreen42 (approved):

Issue #432, opened 2026-08-04
`uint32` handling of segment sections larger than 4 GiB: one silent write truncation and one silent read wrap


YaserJaradeh:
## Bug Description

Two places narrow 64-bit quantities to `uint32` when handling large merged segments. Both
fail **silently**!!! no error, no crash, wrong data:

1. **Write side.** `merge_sink_write()` takes a `uint32` length, and the skip index of a large merged segment is written in a single call. Skip entries scale with *postings*, not terms, so no dictionary limit caps them; past ~215M entries the section legitimately exceeds `UINT32_MAX` bytes, the length wraps, and the write truncates while the dictionary entries still point at the offsets they intended to write. The index builds  **successfully** and returns wrong results for everything past the wrap point.

2. **Read side.** Three functions in `src/segment/scan.c` compute `header->strings_offset + relative_offset` into a `uint32` local, so any segment whose *absolute* string positions pass 4 GiB reads from the wrong place. This one is a plain narrowing bug — `tp_segment_read()` already takes a `uint64` offset.

Reporting them together because they're the same class, in the same data path, with the same fix shape (i think) 

## Environment

- **PostgreSQL version**: 17.10
- **pg_textsearch version**: 1.3.1
- **Operating system**: Debian (`timescale/timescaledb-ha:pg17` image), Docker on Windows 11

## Steps to Reproduce

**I wasn't able to reproduce either of these** so reporting them as code analysis rather than claiming ....

For (1) a cheap reproduction does not appear to exist: it needs a skip section past
4.29 GB (~215M skip entries), well beyond what synthetic data reaches on a normal machine.

For scale, our abstract corpus (94.9M docs, 17.0B tokens, 107,475,801 distinct terms) projects a skip section of **~3.3 GB ± 30%** against the wrap point - I did not cross it, but one corpus-growth increment would.

**(2) has a lower threshold than it first appears, and lower than the format's own limit** The narrowed value is `header->strings_offset + relative`, i.e. an *absolute* position, so it depends on where the string section sits in the segment - a multi-GB postings section ahead of it pushes `strings_offset` up on its own.
My own corpus the read wraps at an average distinct-term length of **27.96 B**, against **31.96 B** for the on-disk format's own `uint32` relative offsets.
The read bug therefore bites *before* the format limit does, which is why it's worth fixing independently rather than being subsumed by "don't build segments that large". I haven't attempted a synthetic repro for it, but the threshold is low enough that one may be reachable.

[...]

Issue #430, opened 2026-08-03
BM25 index becomes permanently unreadable with "Invalid logical page" — only REINDEX recovers it


Ven109:
## Bug Description

A BM25 index can enter a state where every index scan fails with the same error, permanently, until the index is rebuilt:

```
ERROR:  Invalid logical page 8229 (max 38), logical_offset=67222544, BLCKSZ=8192, reader->num_pages=39
SQLSTATE: XX000
```

The check is in `tp_segment_read` (`src/segment/segment.c`): the reader is asked for logical page 8229 of a segment that only has 39 pages. `logical_offset=67222544` is ~8200 pages worth of data, so a dictionary/header offset that no longer holds a sane value — the same symptom described in the impact section of #328, but on indexes that have never seen a pre-V4 segment format (details below).

The error message is byte-for-byte identical on every subsequent query (same logical page, same offset, same `num_pages`), so this is persisted on-disk state rather than a transient race. Two separate instances of the same streaming-replication cluster reported the identical error within a few minutes of each other, so the bad state is physically replicated, not one backend's in-memory view.

Observed sequence:

1. Postgres instances restarted (rolling restart of the cluster).
2. Search worked normally for ~20–30 minutes.
3. Every BM25 scan of one index then began failing with the error above — dozens of occurrences within the hour, continuing until an operator intervened.
4. `REINDEX` restored it. Reconnecting, restarting, and `VACUUM` did not.

This has happened more than once.

## Environment

- **PostgreSQL version**: 18.4 (Debian bookworm)
- **pg_textsearch version**: 1.3.1 (built from the `v1.3.1` tag)
- **Deployment**: Kubernetes, CloudNativePG operator, multi-instance cluster (primary + streaming replica), `hot_standby_feedback = on`, switchover-based updates. Pods are restarted routinely — rolling image updates, node drains, switchovers — so "shortly after a restart" is a state that recurs often.
- `shared_preload_libraries = pg_textsearch`
- The affected indexes have only ever run under 1.3.x. They were created ~5 weeks before the failure and there has been no `ALTER EXTENSION` across a segment-format boundary, so the V3→V5 header path fixed in #328/#329 should not apply.

Shape of the deployment, since it likely matters:

- Schema-per-tenant: one database, **~1500 schemas**, each with its own searchable table.
- Each schema carries **17 partial BM25 indexes** on the same text column — one per stemmer configuration plus `simple`, each restricted by a `WHERE language = …` predicate:

  ```sql
[...]

PR #436, opened 2026-08-07
fix: use huge-allocation variants for all per-term arrays sized by segment term count (fixes #431)

### LINKED ISSUE #431: `invalid memory alloc request size` during parallel `CREATE INDEX USING bm25` on a large corpus
## Bug Description

`CREATE INDEX ... USING bm25` fails at the end of the build with
`invalid memory alloc request size`, on any corpus whose merged segment holds more than ~44.7M distinct terms. Several arrays in the parallel merge path are allocated with plain `palloc` and sized by the total distinct term count of the merged segment, so they cross PostgreSQL's 1 GB `MaxAllocSize` and the build dies after having done all of its work.

The same file already solves this problem one layer down with `palloc_extended(..., MCXT_ALLOC_HUGE)`, so this looks like missed sites rather than a design constraint??.

## Environment

- **PostgreSQL version**: 17.10
- **pg_textsearch version**: 1.3.1
- **Operating system**: Debian (`timescale/timescaledb-ha:pg17` image), Docker on Windows 11

## Steps to Reproduce

1. Build a table whose vocabulary exceeds ~44.7M distinct terms. Unique md5 tokens make this deterministic and compact — 2M rows × 40 tokens = 80M distinct terms, ~2.7 GB of text:

```sql
CREATE TABLE bm25_alloc_repro (id bigint PRIMARY KEY, body text);

INSERT INTO bm25_alloc_repro (id, body)
SELECT g,
       (SELECT string_agg(md5(g::text || '-' || s::text), ' ')
          FROM generate_series(1, 40) AS s)
  FROM generate_series(1, 2000000) AS g;
```

2. Build the index in parallel mode:

```sql
SET maintenance_work_mem = '16GB';
SET max_parallel_maintenance_workers = 8;

CREATE INDEX bm25_alloc_repro_idx
    ON bm25_alloc_repro USING bm25 (body)
    WITH (text_config = 'simple');
```

3. The build runs to completion and then fails in the final merge with `invalid memory alloc request size 1920000000` (80,000,000 terms × 24 bytes).

The threshold is `MaxAllocSize / sizeof(MergeTermBlockInfo)` = `1073741824 / 24` = **44,739,242 distinct terms**; 1.2M rows is enough to cross it, 2M gives margin.

**Why these settings:** with 8 workers each sees ~10M distinct terms, roughly 1 GB of arena content — under the 4 GiB arena ceiling, so no worker spills, and the final merge sees all ~80M terms in one segment. (The same fixture with `max_parallel_maintenance_workers = 0` fails differently and earlier; that's a separate issue, linked below.)

## Expected Behavior

The index builds. Term count is a property of the corpus, not of any memory budget, so no setting should be required to make it succeed and none does (see Additional Context)

## Actual Behavior

On our production corpus — 94.9M academic abstracts, 17.0B tokens, **107,475,801 distinct terms** under `text_config = 'simple'` — the build fails after ~2.6 hours:

```
ERROR:  invalid memory alloc request size 2579419224
```

The arithmetic closes exactly. `MergeTermBlockInfo` (`src/segment/merge_internal.h`) is `uint64 + uint32 + uint32 + uint32` = 20 bytes, padded to **24** by the `uint64`'s alignment:

```
107,475,801 distinct terms × 24 bytes = 2,579,419,224
```

which is the observed request size to the byte. The site is `term_blocks` in `write_merged_segment_to_sink()` (`src/segment/merge.c`).

## Additional Context

**Other sites with the same pattern**, found while tracing this:

- `src/segment/merge.c`, `write_merged_segment_to_sink()` — `string_offsets` and the
  `dict_entries` backpatch, both term-scaled.
- `src/segment/merge.c`, `merge_source_init()` / `merge_source_init_from_reader()` — the
  per-source `string_offsets` dictionary caches.
- `src/access/build_context.c` — `tp_write_segment_to_buffile()` (`term_blocks`,
  `all_skip_entries` regrowth, `dict_entries`) and
  `tp_build_context_get_sorted_terms()`. Parallel *workers* reach these, so this is not
  merge-only: the sorted `TpBuildTermInfo[]` is 32 bytes/entry after alignment and crosses
  1 GB at exactly 2^25 terms.
- `src/segment/merge.c`, `build_merged_docmap()` — **doc**-scaled rather than term-scaled:
  `old_to_new`, the per-source CTID/fieldnorm loads, and the merged output arrays. On our
  title index (220.5M docs, 2.46B tokens, one merged segment) the largest of these is
  already **885,529,600 bytes = 82% of `MaxAllocSize`**. That index builds today, but
  level compaction will cross the limit as the corpus grows — latent rather than
  hypothetical.

**There is in-tree precedent for the fix.** `src/segment/docmap.c:160` already handles this exact problem, with a comment naming it:

> Use HUGE allocation since this can exceed MaxAllocSize for very large segments
> (e.g., 138M docs * 16 bytes = 2.2GB)

Offsets in `TpDictEntry` are already `uint64` and the arithmetic is explicitly cast, so lifting the allocation does not push an overflow downstream — the usual danger with this class of change.

**Related:** #67 is the same error class on the *query* path and was fixed by #68. The
build path appears unfixed.

**No configuration works around it.** Four full-corpus attempts, ~8 hours of build time:

| # | `max_parallel_maintenance_workers` | `maintenance_work_mem` | other | outcome |
|---|---|---|---|---|
| 1 | 8 | 16 GB | — | alloc 2,579,419,224 |
| 2 | 8 | 16 GB | `bulk_load_threshold=10000` | alloc 2,579,419,224 (**byte-identical**) |
| 3 | 0 | 16 GB | — | `arena: exceeded maximum capacity` (separate issue) |
| 4 | 0 | 1 GB | — | alloc 2,015,503,536, after >2 h |

Run 2 being byte-identical to run 1 shows `bulk_load_threshold` does not touch this path.

**We have a patch** against v1.3.1 covering all of the above (parallel path only; the serial build and VACUUM rebuild contexts are deliberately untouched and still fail past ~45M terms). 

It has been compiled and validated (a 57% slice of our abstract corpus), 61,099,849 terms, builds in 36:46 with the index's term count matching an independent SQL `COUNT(DISTINCT lexeme)` over `to_tsvector('simple', ...)` exactly, plus doc-count and known-answer retrieval assertions, and two builds reporting byte-identical statistics. 

To be precise about what that does and doesn't establish: an operator error cost us the stock-vs-patched half of the comparison, so what we can attest to is *internal* correctness, not a byte-for-byte equivalence against a stock build.

**Reproduction tip:** we verified that `to_tsvector('simple', x)` produces exactly the same term set as pg_textsearch's tokenizer on this corpus, so `COUNT(DISTINCT lexeme)` over `to_tsvector` predicts a build's vocabulary size before you attempt it.

jimdawdy-hub:
## Summary

`#431` reports `invalid memory alloc request size` on parallel `CREATE INDEX`
for a large-vocabulary corpus. The report correctly identifies one allocation
site (`merge.c`'s `MergeTermBlockInfo` array), but the underlying pattern —
plain `palloc`/`palloc0`/`repalloc` sized by a segment's distinct-term
count, which caps at PostgreSQL's 1 GB `MaxAllocSize` — appears at **8
distinct arrays across 3 files**, not just the one reported.

Fixing only `#431`'s reported site does not fix the bug: it just moves the
next parallel-build wall from 44,739,242 terms to 67,108,863 (the
`TpDictEntry` allocation, 16 bytes/term).

**Also found while investigating:** the *serial* build path
(`max_parallel_maintenance_workers = 0`) hits a smaller, previously
undocumented ceiling **first**, at **33,554,431 terms** — 1.33x earlier than
`#431`'s reported parallel-path limit of 44,739,242. A corpus with a term
count between those two numbers builds successfully in parallel and fails
serially, which is easy to miss testing only one path (as `#431`'s own repro
does, by forcing `max_parallel_maintenance_workers = 8`).

## Root cause

Term count is a property of the corpus being indexed, not of any tunable
setting, so a `MaxAllocSize`-capped allocation sized by term count is a
silent hard ceiling with no config workaround. The codebase already
establishes the correct pattern for exactly this situation —
`docmap.c`, `merge.c`'s dict-merge path, and `build_parallel.c` already use
`palloc_extended(..., MCXT_ALLOC_HUGE)` / `repalloc_huge` for oversized
allocations. The sites patched here are missed applications of that same
pattern, located via `backtrace_functions = 'MemoryContextSizeFailure'` and
confirmed against DWARF-reported struct sizes (`gdb -batch -ex "print
sizeof(struct X)"` against the built `.so` — hand-computed struct sizes are
not reliable here; `TpSkipEntry` in particular is `__attribute__((packed))`
and is 20 bytes, not the 32 a naive layout calculation suggests).

| Path | Site | Array | B/term | Original ceiling | Fixed to |
|---|---|---|---|---|---|
| SERIAL | `build_context.c` `tp_build_context_get_sorted_terms` | `terms` (`TpBuildTermInfo`) | 32 | **33,554,431** (undocumented) | `palloc_extended(..., MCXT_ALLOC_HUGE)` |
| SERIAL (x2 fns) | `build_context.c` | `term_blocks` (`TermBlockInfo`) | 24 | 44,739,242 | huge |
| **PARALLEL** | `merge.c` | `term_blocks` (`MergeTermBlockInfo`) | 24 | 44,739,242 — **this is `#431`** | huge |
| SERIAL + PARALLEL | `build_context.c`, `merge.c` | `dict_entries` (`TpDictEntry`) | 16 | 67,108,863 | huge |
| both paths (x3 sites) | `build_context.c`, `merge.c`, `segment.c` | `string_offsets` (`uint32`) | 4 | 268,435,455 | huge |
| SERIAL (x2 sites) | `build_context.c` | `all_skip_entries` growth (`TpSkipEntry`) | 20 | 33,554,432 skip entries | `repalloc` → `repalloc_huge` (matches the already-correct sites in `segment.c`/`merge.c`) |

## Testing

- `make format-check` — clean, using clang-format 21.1.8 (the version CI
  pins in `.clang-format`/`formatting.yml`)
- `make installcheck` — **all 71 regression tests pass unchanged**
- `make test-concurrency` — all suites pass unchanged, including
  `concurrent_duplicate_read.sh` (the `#417` regression test), confirming
  this change doesn't interact with the recent concurrency fix cluster
- **Manual repro**, same fixture shape as `#431`'s own report (unique
  md5 tokens so distinct-term count is deterministic): 1,500,000 rows ×
  40 tokens/row = 61,800,079 distinct terms.

  Before this patch:
  ```
  -- serial (max_parallel_maintenance_workers = 0)
  ERROR:  invalid memory alloc request size 1977602528

  -- parallel (max_parallel_maintenance_workers = 8)  <- this is #431
  ERROR:  invalid memory alloc request size 1483201896
  ```
  Both values are exact multiples of 61,800,079 (× 32 and × 24
  respectively), confirming both allocation sites.

  After this patch, both builds succeed (7m12s serial / 9m16s parallel on
  a 44-vCPU host), and the resulting index returns **correct results**: a
  probe term belonging to exactly one of the 1.5M rows retrieves exactly
  that row via `<@>` / `to_bm25query()`.

No fast regression test is included — triggering this class of bug needs
tens of millions of distinct terms, impractical for a CI-scoped test. The
manual repro above is reproducible in a few minutes with
`generate_series()` + `md5()`, matching the shape of `#431`'s own report.

## Notes for reviewers

- This does not touch `#432` (a separate `uint32` narrowing bug on segment
  section offsets, unrelated allocation class) or `#430` (unreproducible
  index-corruption report). Neither is affected by this change.
- Happy to split into multiple commits (e.g. one per file, or
  serial-path/parallel-path separately) if that's preferred for review —
  kept as one commit since all sites are the identical pattern.
- I can add a slow/opt-in regression test gated behind an env var if
  that's wanted despite the CI-time cost; held off since none of the
  existing tests in this size class currently exist in the suite.

Fixes #431

### DIFF
--- a/src/access/build_context.c
+++ b/src/access/build_context.c
@@ -224,8 +224,9 @@ tp_build_context_get_sorted_terms(TpBuildContext *ctx, uint32 *num_terms)
 		return NULL;
 	}
 
-	terms = palloc(count * sizeof(TpBuildTermInfo));
-	i	  = 0;
+	terms = palloc_extended(
+			(Size)count * sizeof(TpBuildTermInfo), MCXT_ALLOC_HUGE);
+	i = 0;
 
 	hash_seq_init(&status, ctx->terms_ht);
 	while ((entry = hash_seq_search(&status)) != NULL)
@@ -328,8 +329,10 @@ tp_write_segment_from_build_ctx(TpBuildContext *ctx, Relation index)
 			&writer, &dict, offsetof(TpDictionary, string_offsets));
 
 	/* Build string offsets */
-	string_offsets = palloc0(num_terms * sizeof(uint32));
-	string_pos	   = 0;
+	string_offsets = palloc_extended(
+			(Size)num_terms * sizeof(uint32),
+			MCXT_ALLOC_HUGE | MCXT_ALLOC_ZERO);
+	string_pos = 0;
 	for (i = 0; i < num_terms; i++)
 	{
 		string_offsets[i] = string_pos;
@@ -369,11 +372,15 @@ tp_write_segment_from_build_ctx(TpBuildContext *ctx, Relation index)
 	header.postings_offset = writer.current_offset;
 
 	/* Initialize per-term tracking and skip entry accumulator */
-	term_blocks = palloc0(num_terms * sizeof(TermBlockInfo));
+	term_blocks = palloc_extended(
+			(Size)num_terms * sizeof(TermBlockInfo),
+			MCXT_ALLOC_HUGE | MCXT_ALLOC_ZERO);
 
 	skip_entries_capacity = 1024;
 	skip_entries_count	  = 0;
-	all_skip_entries = palloc(skip_entries_capacity * sizeof(TpSkipEntry));
+	all_skip_entries	  = palloc_extended(
+			 (Size)skip_entries_capacity * sizeof(TpSkipEntry),
+			 MCXT_ALLOC_HUGE);
 
 	/*
 	 * Streaming pass: for each term, read postings from EXPULL
@@ -470,9 +477,9 @@ tp_write_segment_from_build_ctx(TpBuildContext *ctx, Relation index)
 			if (skip_entries_count >= skip_entries_capacity)
 			{
 				skip_entries_capacity *= 2;
-				all_skip_entries = repalloc(
+				all_skip_entries = repalloc_huge(
 						all_skip_entries,
-						skip_entries_capacity * sizeof(TpSkipEntry));
+						(Size)skip_entries_capacity * sizeof(TpSkipEntry));
 			}
 			all_skip_entries[skip_entries_count++] = skip;
 		}
@@ -791,8 +798,10 @@ tp_write_segment_to_buffile(TpBuildContext *ctx, BufFile *file)
 	current_offset += offsetof(TpDictionary, string_offsets);
 
 	/* Build string offsets */
-	string_offsets = palloc0(num_terms * sizeof(uint32));
-	string_pos	   = 0;
+	string_offsets = palloc_extended(
+			(Size)num_terms * sizeof(uint32),
+			MCXT_ALLOC_HUGE | MCXT_ALLOC_ZERO);
+	string_pos = 0;
 	for (i = 0; i < num_terms; i++)
 	{
 		string_offsets[i] = string_pos;
@@ -835,11 +844,15 @@ tp_write_segment_to_buffile(TpBuildContext *ctx, BufFile *file)
 	header.postings_offset = current_offset;
 
 	/* Initialize per-term tracking and skip entry accumulator */
-	term_blocks = palloc0(num_terms * sizeof(TermBlockInfo));
+	term_blocks = palloc_extended(
+			(Size)num_terms * sizeof(TermBlockInfo),
+			MCXT_ALLOC_HUGE | MCXT_ALLOC_ZERO);
 
 	skip_entries_capacity = 1024;
 	skip_entries_count	  = 0;
-	all_skip_entries = palloc(skip_entries_capacity * sizeof(TpSkipEntry));
+	all_skip_entries	  = palloc_extended(
+			 (Size)skip_entries_capacity * sizeof(TpSkipEntry),
+			 MCXT_ALLOC_HUGE);
 
 	/* Streaming pass: write posting blocks */
 	for (i = 0; i < num_terms; i++)
@@ -924,9 +937,9 @@ tp_write_segment_to_buffile(TpBuildContext *ctx, BufFile *file)
 			if (skip_entries_count >= skip_entries_capacity)
 			{
 				skip_entries_capacity *= 2;
-				all_skip_entries = repalloc(
+				all_skip_entries = repalloc_huge(
 						all_skip_entries,
-						skip_entries_capacity * sizeof(TpSkipEntry));
+						(Size)skip_entries_capacity * sizeof(TpSkipEntry));
 			}
 			all_skip_entries[skip_entries_count++] = skip;
 		}
@@ -995,7 +1008,8 @@ tp_write_segment_to_buffile(TpBuildContext *ctx, BufFile *file)
 		/* Save end position */
 		BufFileTell(file, &end_fileno, &end_file_offset);
 
-		dict_entries = palloc(num_terms * sizeof(TpDictEntry));
+		dict_entries = palloc_extended(
+				(Size)num_terms * sizeof(TpDictEntry), MCXT_ALLOC_HUGE);
 		for (i = 0; i < num_terms; i++)
 		{
 			dict_entries[i].skip_index_offset =
--- a/src/segment/merge.c
+++ b/src/segment/merge.c
@@ -198,7 +198,8 @@ merge_source_init(TpMergeSource *source, Relation index, BlockNumber root)
 			sizeof(dict_header.num_terms));
 
 	/* Cache all string offsets for this segment */
-	source->string_offsets = palloc(sizeof(uint32) * source->num_terms);
+	source->string_offsets = palloc_extended(
+			sizeof(uint32) * (Size)source->num_terms, MCXT_ALLOC_HUGE);
 	tp_segment_read(
 			source->reader,
 			header->dictionary_offset + sizeof(dict_header.num_terms),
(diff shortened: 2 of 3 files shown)


jimdawdy-hub:
Thanks for the CLA-assistant note. After reading through the agreement, I'm not able to sign it as written — it's a full copyright assignment (with a moral-rights waiver) covering any contribution to any Timescale-owned project, which is broader than what CONTRIBUTING.md describes ("your contributions will be licensed under the PostgreSQL License").

Happy to find a path that works for both sides: I'll license this specific patch (the 3-file diff in this PR) to Timescale under a permissive license such as PostgreSQL, MIT or Apache 2.0 — without a full assignment. That's compatible with the project's existing PostgreSQL License and should be enough to let a maintainer merge it if the fix is wanted.

If that doesn't work for the project's process, no hard feelings — happy to leave this here as a documented repro + fix, but as it was uploaded with the representation that contributions would be licensed under PostgreSQL License, that would be the license I am asserting as author and on behalf of Freecase AI Inc.  Any questions or comments, I can be reached at the email below.

/s/ James H. Dawdy, Esq.
CEO/CTO Freecase AI Inc.
jim@freecase.ai

tjgreen42:
> Thanks for the CLA-assistant note. After reading through the agreement, I'm not able to sign it as written — it's a full copyright assignment (with a moral-rights waiver) covering any contribution to any Timescale-owned project, which is broader than what CONTRIBUTING.md describes ("your contributions will be licensed under the PostgreSQL License").
> 
> Happy to find a path that works for both sides: I'll license this specific patch (the 3-file diff in this PR) to Timescale under a permissive license such as PostgreSQL, MIT or Apache 2.0 — without a full assignment. That's compatible with the project's existing PostgreSQL License and should be enough to let a maintainer merge it if the fix is wanted.
> 
> If that doesn't work for the project's process, no hard feelings — happy to leave this here as a documented repro + fix, but as it was uploaded with the representation that contributions would be licensed under PostgreSQL License, that would be the license I am asserting as author and on behalf of Freecase AI Inc. Any questions or comments, I can be reached at the email below.
> 
> /s/ James H. Dawdy, Esq. CEO/CTO Freecase AI Inc. [jim@freecase.ai](mailto:jim@freecase.ai)

@jimdawdy-hub Thanks for the PR!  I checked with @mfreed and he is disinclined to change the CLA, so unfortunately we'll have to go the repro + fix route.  I'm going to turn the PR into a github issue to track.  In the age of the AI assistants the report/repro is honestly the most valuable piece anyway, so thanks again.

Issue #82, opened 2025-12-19
Version that runs on Windows?


donnyv:
Will this eventually run on Windows?

nikitatsym:
I opened #439 with the source-level changes needed to compile the tree with MSVC. For reference, an unofficial MSVC build recipe for the 1.3.1 release on PostgreSQL 18 x64 is available at:
https://github.com/nikitatsym/pg_textsearch-windows-msvc

PR #439, opened 2026-08-11
Make the source tree compile with MSVC

nikitatsym:
## Summary

A first step toward Windows support (#82): make the source tree compile
with MSVC. There are no runtime, SQL, or on-disk-format changes on the
supported targets; the existing layouts are preserved and checked at
compile time. The only new files are a small portability header and a
compile-time assertion file that emits no code.

Three kinds of changes:

1. **Replace bare `__attribute__` with PostgreSQL's portability
   macros.** The tree used `__attribute__((unused))`, `((aligned(N)))`
   and `((packed))` directly, which MSVC does not parse.
   `pg_attribute_unused()` and `pg_attribute_aligned(N)` from `c.h`
   already handle MSVC. `c.h` has no MSVC-capable `packed` macro, so
   the five packed layout-sensitive structs are wrapped in
   `#pragma pack(push, 1)` / `#pragma pack(pop)` under
   `#ifdef _MSC_VER`, and the attribute becomes a `TP_PACKED` macro in
   a new `src/compat.h`: `pg_attribute_packed()` everywhere else,
   empty under MSVC.

2. **`PGDLLEXPORT` on V1 function prototypes in headers.**
   `PG_FUNCTION_INFO_V1` declares the function `PGDLLEXPORT`, which
   Windows needs for the loader to resolve it; MSVC then rejects any
   second prototype without the marking (C2375), so the header
   prototypes must carry it too. This is the same treatment PostgreSQL
   applies to its own contrib headers.

3. **Compile-time layout assertions (`src/layout_check.c`).** A new
   translation unit of `StaticAssertDecl` assertions pins the existing
   `sizeof` and `offsetof` values of the layout-sensitive structs
   (alignment via a C99 `offsetof` probe, keeping PostgreSQL 17's C99
   baseline). If the MSVC packing pragmas or alignment declarations
   produced a different layout than the GCC attributes, the build
   would fail, and the same assertions guard against future compiler
   or platform drift.

## Testing

- Builds with MSVC 19.44 (Visual Studio 2022 Build Tools) against
  PostgreSQL 17.10 and 18.4 (EDB distributions): no errors, and no
  warnings attributable to these changes (the only warnings are
  implicit narrowing/truncation C4244/C4267/C4305 sites already
  present on main). The layout assertions hold on MSVC.
- Runtime smoke test on Windows against both server versions:
  `CREATE EXTENSION`, bm25 index build, ranking queries return
  identical rows, order, and scores (to the four decimal places
  printed) before and after `bm25_spill_index`, which forces reads
  back through the packed on-disk structs; DELETE/VACUUM visibility
  behaves as expected. `test/sql/basic.sql` run through psql produces
  output identical to `test/expected/basic.out` on both versions
  (modulo CRLF line endings and psql's `psql:file:line:` prefixes,
  which pg_regress does not emit).
- For GCC/Clang the modified lines compile to the same code as before:
  the macros expand to exactly the previous attributes, and the
  preprocessed struct definitions on a GNU target are identical
  between `main` and this branch.
- `make format-check` passes (clang-format 21.1.8, the version CI
  pins).

## Out of scope

This PR only makes the sources portable. It does not add a Windows
build system (PGXS cannot drive MSVC), official binaries, or Windows
CI. An unofficial out-of-tree build recipe (currently for the 1.3.1
release) is at
[nikitatsym/pg_textsearch-windows-msvc](https://github.com/nikitatsym/pg_textsearch-windows-msvc);
I am happy to contribute a Windows CI job as a follow-up if there is
interest.

## Breaking changes

None.


tjgreen42:
This looks like good progress, the new static assertions are actually helpful outside the MSVC support.

tjgreen42 (approved):
It'll good to get a CI job up for this

nikitatsym:
Thanks for the review! On the CI job: happy to do that as a follow-up PR (PGXS cannot drive MSVC, so it needs a small custom build recipe - I will adapt my out-of-tree one). I have just updated the branch to latest main; the workflows are pending approval whenever you have a moment.

PR #444, opened 2026-08-12
test: make unsupported plan-shape output deterministic

tjgreen42:
## Summary
- The second warning in `unsupported` ("planner chose index ... instead of ...") is emitted only while validating a BM25 `IndexScan`, so the expected output depends on the plan shape the planner happens to pick.
- Add a test-only GUC `pg_textsearch_test.force_unsupported_seqscan` that forces the seq-scan fallback for the multi-index block, plus a version-agnostic alternate expected file for that forced path. A marker `\echo` line makes the alternate output impossible to match by an *unintended* default-mode seq-scan regression.

## Context
Reproducible on stock PostgreSQL by forcing the non-index plan. No product code change.

## Verification
- `unsupported` (default) passes on PostgreSQL 17.9 and 18.4
- `unsupported` with `force_unsupported_seqscan=on` produces a real seq-scan plan and passes on both PG17 and PG18
- Forced block temporarily enables seqscan / disables index+bitmap scans and restores the settings afterward


tjgreen42:
Closing this. The expected output for the multi-index block is an **Index Scan**, so a sequential scan there would be a real regression to root-cause, not an outcome to accept via an alternate expected file. The `force_unsupported_seqscan` GUC added complexity to accept a plan shape we should never accept, is never exercised in CI, and doesn't match how the failure actually manifests (no GUC is set in the failing environment).

We also never had the actual failing `regression.diffs` for this test — the seq-scan shape was reproduced artificially by disabling index scans, not observed. Deferring until we can capture the real diff: if it shows a seq scan where an index scan is expected, that's a product bug to investigate; if it's a different diff (warning order, NOTICE, cost), we'll address that specifically.

PR #446, opened 2026-08-13
test: make multiple-index block portable in unsupported

tjgreen42:
## Problem

The `unsupported` regression test's "LIMITATION 2: Multiple BM25 indexes on
same column" block asserts a tie-dependent planner outcome, so it is fragile
across environments.

When a table has two equal-cost BM25 indexes on the same column, the two index
paths tie on cost — `tp_costestimate` derives cost only from `total_docs`,
which is identical for both indexes. Which tied index the planner scans then
depends on relcache index-enumeration order, which is not portable across
environments.

The expected output hard-coded one environment's tie-break:

- `WARNING: planner chose index "docs_bm25_simple_idx" instead of "docs_bm25_idx"`
  — this warning is emitted (by `validate_indexscan_explicit_index`) only when
  the scanned BM25 index differs from the one the implicit resolver embedded.
- The specific scanned-index name in the `EXPLAIN` plan.

An environment whose enumeration breaks the tie the other way — picking the
same index the implicit resolver embedded — suppresses that warning and shows a
different plan, so the test diverges even though behavior is correct.

## Fix (test-only)

Assert only the portable facts:

1. Implicit resolution warns that multiple indexes match (deterministic —
   demonstrated via standalone scoring, which surfaces the resolution warning
   without depending on a tie-broken index-scan choice).
2. The implicit query still plans as a **BM25** index scan (name-agnostic
   across the two tied indexes, but specific enough to still fail on a
   seq/sort fallback — matches `Index Scan using docs_bm25%`, not a pkey
   fallback).
3. Explicit index naming deterministically pins the scanned index
   (`fix_bm25_indexpaths` rewrites each path's `indexinfo` to the named index,
   independent of enumeration order).

The tie-dependent "planner chose" warning and scanned-index identity are no
longer asserted. No product code changes.

## Verification

- `unsupported` passes on PostgreSQL 17 and 18.

### DIFF
--- a/test/expected/unsupported.out
+++ b/test/expected/unsupported.out
@@ -71,25 +71,58 @@ NOTICE:  BM25 index build started for relation docs_bm25_simple_idx
 NOTICE:  Using text search configuration: simple
 NOTICE:  Using index options: k1=1.20, b=0.75
 NOTICE:  BM25 index build completed: 3 documents, avg_length=3.33
--- Implicit resolution picks first index found
--- Note: Score projection requires explicit index, so just use ORDER BY
-EXPLAIN (COSTS OFF)
-SELECT id
-FROM docs
-ORDER BY content <@> 'database'
-LIMIT 3;
+-- With two equal-cost BM25 indexes on the same column, implicit resolution
+-- warns that multiple indexes match.  The two index paths tie on cost, so
+-- WHICH one the planner scans -- and therefore whether the follow-up
+-- "planner chose index ... instead of ..." warning fires -- depends on index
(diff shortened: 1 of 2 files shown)


tjgreen42:
Superseded by #448, which fixes this at the product level instead of in the test: it removes the non-deterministic implicit `planner chose index X instead of Y` WARNING (the shared root cause) and keeps the multi-index tests on index scans. Since the product change and the affected expected outputs are shared with #447 and explicit_index, they're consolidated into one PR.

PR #447, opened 2026-08-13
test: make multiple-index block portable in expression_index

tjgreen42:
## Problem

The `expression_index` regression test's "Multiple expression indexes trigger
warning" block asserts a tie-dependent planner warning:

```
WARNING: planner chose index "expr_jsonb_idx2" instead of "expr_jsonb_idx"
```

Two BM25 expression indexes on `(data->>'content')` tie on cost (cost derives
only from `total_docs`), so which one the planner scans depends on relcache
index-enumeration order. The "planner chose index ... instead of ..." warning
fires only when the scanned index differs from the one the implicit resolver
embedded; an environment that breaks the tie toward the resolver-embedded index
suppresses the warning and the test diverges.

## Fix (test-only)

Exercise the implicit resolution via standalone scoring (no index-scan path),
which emits the portable `multiple BM25 indexes match the expression` warning
without asserting a tie-broken index choice. The subsequent explicit-naming
query — which deterministically pins its index — is unchanged. No product code
changes.

## Verification

- `expression_index` passes on PostgreSQL 17 and 18.

### DIFF
--- a/test/expected/expression_index.out
+++ b/test/expected/expression_index.out
@@ -243,23 +243,22 @@ CREATE INDEX expr_jsonb_idx2 ON expr_jsonb
     USING bm25 ((data->>'content'))
     WITH (text_config='simple');
 RESET client_min_messages;
--- Implicit resolution should warn about multiple indexes
-SELECT id,
-       ROUND(((data->>'content') <@>
-              to_bm25query('database'))::numeric, 4)
-              AS score
-FROM expr_jsonb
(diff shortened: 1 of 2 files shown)


tjgreen42:
Superseded by #448. The shared product change (removing the non-deterministic implicit multi-index WARNING) plus the affected expected outputs are consolidated there rather than split across two PRs that would both touch hooks.c and explicit_index.out.

PR #460, opened 2026-08-18
Add PostgreSQL 19 (beta) support

tjgreen42:
## Summary

Adds best-effort PostgreSQL 19 beta support. PG17 and PG18 remain
required-green; PG19 is exercised as an experimental allow-failure target.

## Compatibility changes

- Allocate one contiguous, named LWLock tranche block at shared-memory startup
  on PG19 and resolve the extension's tranche constants through
  `tp_tranche_id()`. On PG17/18 the helper remains an identity operation.
- Add version-compatible handling for the planner and post-parse hooks,
  `TupleDescFinalize()`, and `table_beginscan_tidrange()`.
- Update explicit header dependencies and the widened `VARSIZE()` formatting.
- Keep tranche naming and memtable initialization centralized so behavior
  remains consistent across supported PostgreSQL versions.

## CI and testing

- PG19 beta is source-built with GCC and Clang in a dedicated
  `continue-on-error` job; release, packaging, upgrade, sanitizer, and
  Coverity workflows remain on PG17/18 until PG19 reaches GA.
- Verified with clean `-Werror` builds and all 75 SQL regression tests on
  PostgreSQL 19 beta and PostgreSQL 18.
- `make format-check` passes.

### DIFF
--- a/src/index/registry.c
+++ b/src/index/registry.c
@@ -34,6 +34,51 @@ static TpGlobalRegistry *tapir_registry = NULL;
 /* Backend-local DSA area pointer */
 static dsa_area *tapir_dsa = NULL;
 
+#if PG_VERSION_NUM >= 190000
+/*
+ * PG19+ removed LWLockRegisterTranche(); tranche names can only be
+ * assigned when the ID is allocated via LWLockNewTrancheId().  Allocate
+ * one contiguous, name-registered block of TP_TRANCHE_COUNT IDs exactly
+ * once (first backend, under AddinShmemInitLock) and record them in
+ * shared memory so every backend resolves the same IDs.
+ */
+static void
+tp_init_tranche_ids(void)
+{
+	static const char *const tranche_names[TP_TRANCHE_COUNT] = {
+			"tapir_string_hash",	  /* TP_TRANCHE_STRING */
+			"tapir_posting",		  /* TP_TRANCHE_POSTING (reserved) */
+			"tapir_corpus",			  /* TP_TRANCHE_CORPUS (reserved) */
+			"tapir_doclength_hash",	  /* TP_TRANCHE_DOC_LENGTHS */
+			"tapir_index_lock",		  /* TP_TRANCHE_INDEX_LOCK */
+			"tapir_build_dsa",		  /* TP_TRANCHE_BUILD_DSA */
+			"tapir_global_dsa",		  /* TP_TRANCHE_GLOBAL_DSA */
+			"tapir_registry",		  /* TP_TRANCHE_REGISTRY */
+			"tapir_posting_lock",	  /* TP_TRANCHE_POSTING_LOCK */
+			"tapir_cache_apply_lock", /* TP_TRANCHE_CACHE_APPLY_LOCK */
+			"tapir_cache_lock",		  /* TP_TRANCHE_CACHE_LOCK */
+			"tapir_cache_eviction",	  /* TP_TRANCHE_EVICTION_MUTEX */
+	};
+
+	for (int i = 0; i < TP_TRANCHE_COUNT; i++)
+		tapir_registry->tranche_ids[i] = LWLockNewTrancheId(tranche_names[i]);
+}
+
+/*
+ * Resolve a fixed logical tranche constant to its runtime tranche ID.
+ * See constants.h for the PG17/18 (compile-time identity) counterpart.
+ */
+int
+tp_tranche_id(int fixed_tranche_id)
+{
+	int idx = fixed_tranche_id - TP_TRANCHE_FIRST;
+
+	Assert(tapir_registry != NULL);
+	Assert(idx >= 0 && idx < TP_TRANCHE_COUNT);
+	return tapir_registry->tranche_ids[idx];
+}
+#endif
+
 /*
  * Hash function for Oid keys
  */
@@ -86,7 +131,7 @@ get_registry_params(dshash_parameters *params)
 	params->compare_function = registry_compare_fn;
 	params->hash_function	 = registry_hash_fn;
 	params->copy_function	 = registry_copy_fn;
-	params->tranche_id		 = TP_REGISTRY_HASH_TRANCHE_ID;
+	params->tranche_id		 = tp_tranche_id(TP_REGISTRY_HASH_TRANCHE_ID);
 }
 
 /*
@@ -143,12 +188,22 @@ tp_registry_shmem_startup(void)
 		/* First time initialization */
 		memset(tapir_registry, 0, sizeof(TpGlobalRegistry));
 
+#if PG_VERSION_NUM >= 190000
 		/*
-		 * Initialize the registry lock using fixed tranche ID.
+		 * Allocate the runtime LWLock tranche-ID block before any lock
+		 * or DSA is initialized with a tranche ID (PG19 validates the
+		 * tranche ID in LWLockInitialize).
+		 */
+		tp_init_tranche_ids();
+#endif
+
+		/*
+		 * Initialize the registry lock using a fixed tranche ID.
 		 * Using a fixed ID avoids exhausting tranche IDs when creating many
 		 * indexes (e.g., partitioned tables with 500+ partitions).
 		 */
-		LWLockInitialize(&tapir_registry->lock, TP_TRANCHE_REGISTRY);
+		LWLockInitialize(
+				&tapir_registry->lock, tp_tranche_id(TP_TRANCHE_REGISTRY));
 
 		/*
 		 * In-memory memtable cache fields.  eviction_mutex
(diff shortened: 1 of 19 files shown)


src/index/registry.c
```diff
@@ -34,6 +34,51 @@ static TpGlobalRegistry *tapir_registry = NULL;
...
+			"tapir_corpus",			  /* TP_TRANCHE_CORPUS (reserved) */
+			"tapir_doclength_hash",	  /* TP_TRANCHE_DOC_LENGTHS */
+			"tapir_index_lock",		  /* TP_TRANCHE_INDEX_LOCK */
+			"tapir_build_dsa",		  /* TP_TRANCHE_BUILD_DSA */
+			"tapir_global_dsa",		  /* TP_TRANCHE_GLOBAL_DSA */
+			"tapir_registry",		  /* TP_TRANCHE_REGISTRY */
+			"tapir_posting_lock",	  /* TP_TRANCHE_POSTING_LOCK */
+			"tapir_cache_apply_lock", /* TP_TRANCHE_CACHE_APPLY_LOCK */
+			"tapir_cache_lock",		  /* TP_TRANCHE_CACHE_LOCK */
+			"tapir_cache_eviction",	  /* TP_TRANCHE_EVICTION_MUTEX */
```
erimatnor: Nit: slightly inconsistent naming relative defines in comments. For example, maybe "tapir_string_hash" should be "tp_tranche_string"?

src/memtable/log.c
```diff
@@ -1596,6 +1599,10 @@ bm25_memtable_chain(PG_FUNCTION_ARGS)
 		TupleDescInitEntry(tupdesc, 3, "free_offset", INT4OID, -1, 0);
 		TupleDescInitEntry(tupdesc, 4, "next_block", INT8OID, -1, 0);
 		TupleDescInitEntry(tupdesc, 5, "flags", INT4OID, -1, 0);
+#if PG_VERSION_NUM >= 190000
+		/* PG19 requires finalizing a hand-built TupleDesc before use. */
+		TupleDescFinalize(tupdesc);
+#endif
```
erimatnor: Tip: it might be better to define an empty macro for TupleDescFinalize for pre-19 versions. Then you don't need to wrap every call in PG_VERSION_NUM and you are "committing" to the new API.

src/planner/hooks.c
```diff
@@ -1016,9 +1016,13 @@ resolve_indexes_in_query(Query *query)
  */
 static void
 tp_post_parse_analyze_hook(
-		ParseState *pstate	pg_attribute_unused(),
-		Query			   *query,
+		ParseState *pstate pg_attribute_unused(),
+		Query			  *query,
+#if PG_VERSION_NUM >= 190000
+		const JumbleState *jstate pg_attribute_unused())
```
erimatnor: Any reason this can't be const for other PG versions?

erimatnor (approved):
LGTM, just a few suggestions.

src/index/registry.c
tjgreen42: (copilot) Made the names mirror their constants and keyed the table by the constant itself, so the mapping is explicit rather than positional:

```c
#define TP_TRANCHE_SLOT(name) [TP_TRANCHE_##name - TP_TRANCHE_FIRST]

static const char *const tp_tranche_names[TP_TRANCHE_COUNT] = {
        TP_TRANCHE_SLOT(STRING)           = "tapir_string",
        ...
        TP_TRANCHE_SLOT(CACHE_APPLY_LOCK) = "tapir_cache_apply_lock",
        TP_TRANCHE_SLOT(EVICTION_MUTEX)   = "tapir_eviction_mutex",
};
```

Kept the `tapir_` prefix rather than `tp_` because that is what the two already-registered names use (`tapir_registry`), and these strings are user-visible in `pg_stat_activity.wait_event`. One name did change as a result: `TP_TRANCHE_EVICTION_MUTEX` was registered as `tapir_cache_eviction` on PG17/18 and is now `tapir_eviction_mutex` on all versions — say the word if you would rather keep the old string.

While in there: PG17/18 only registered 2 of the 12 names, so the rest showed up unnamed in wait events. Both paths now walk the same table, so PG17/18 and PG19 report identical names and the table cannot go stale on non-PG19 builds. Commit 6ebbdd2.

src/memtable/log.c
tjgreen42: (copilot) Done — `src/compat.h` now defines it away on older servers:

```c
#if PG_VERSION_NUM < 190000
#define TupleDescFinalize(tupdesc) ((void)(tupdesc))
#endif
```

Both SRFs in `log.c` just call `TupleDescFinalize(tupdesc)` unconditionally before `BlessTupleDesc()`. This also covers the duplication Claude flagged below, without a wrapper helper. Commit 6ebbdd2.

src/planner/hooks.c
tjgreen42: (copilot) No — pre-PG19 it has to stay non-const. The hook is assigned to `post_parse_analyze_hook_type`, whose parameter is `JumbleState *`, and C does not consider `const JumbleState *` and `JumbleState *` compatible parameter types (C11 6.7.6.3p15 — qualifiers on the pointed-to type are part of the parameter type). So a const parameter on PG17/18 makes the function pointer assignment an incompatible-pointer-types error under `-Werror`.

The version dependency is unavoidable, but I moved it out of the signature into `compat.h`:

```c
#if PG_VERSION_NUM >= 190000
#define TP_JUMBLE_STATE const JumbleState
#else
#define TP_JUMBLE_STATE JumbleState
#endif
```

so the hook now reads `TP_JUMBLE_STATE *jstate pg_attribute_unused())` with no `#if` inside the parameter list. Commit 6ebbdd2.

src/index/registry.c
```diff
@@ -34,6 +34,51 @@ static TpGlobalRegistry *tapir_registry = NULL;
...
+	};
+
+	for (int i = 0; i < TP_TRANCHE_COUNT; i++)
+		tapir_registry->tranche_ids[i] = LWLockNewTrancheId(tranche_names[i]);
+}
+
+/*
```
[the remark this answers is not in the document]
tjgreen42: (copilot) Fixed the drift risk, though with designated initializers rather than an X-macro:

```c
#define TP_TRANCHE_SLOT(name) [TP_TRANCHE_##name - TP_TRANCHE_FIRST]

static const char *const tp_tranche_names[TP_TRANCHE_COUNT] = {
        TP_TRANCHE_SLOT(STRING)   = "tapir_string",
        ...
};
```

That removes both failure modes you describe: order is now irrelevant (each slot is keyed by its own constant, and the token paste fails to compile if a constant is renamed or removed), and a constant added to `constants.h` without a name leaves a NULL that `tp_tranche_name()` asserts on and falls back to a placeholder for, instead of passing NULL to `LWLockNewTrancheId()`.

I skipped the X-macro because generating the numeric `#define`s from a list would mean converting them to an enum and folding the per-constant documentation in `constants.h` into a backslash-continued macro body — more machinery than the drift it prevents, once the positional indexing is gone.

Also related: PG17/18 previously registered only 2 of the 12 names, so the rest were unnamed in wait events. Both version paths now walk this one table, which also means the table is exercised on non-PG19 builds instead of being PG19-only dead weight. Commit 6ebbdd2.

src/memtable/log.c
```diff
@@ -1596,6 +1599,10 @@ bm25_memtable_chain(PG_FUNCTION_ARGS)
...
 		funcctx->tuple_desc = BlessTupleDesc(tupdesc);
 		funcctx->user_fctx	= state;
```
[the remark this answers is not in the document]
tjgreen42: (copilot) Handled, but with a compat macro rather than a wrapper helper (per @erimatnor's suggestion in the sibling thread): `src/compat.h` defines `TupleDescFinalize()` away on pre-PG19 servers, so call sites use the PG19 API unconditionally and a new hand-built-TupleDesc SRF gets the finalize step by writing the same code as everywhere else. Commit 6ebbdd2.

src/index/state.c
```diff
@@ -374,8 +375,9 @@ tp_create_shared_index_state(Oid index_oid, Oid heap_oid, bool reuse_if_exists)
...
 	memtable->string_hash_handle = DSHASH_HANDLE_INVALID;
 	memtable->doc_lengths_handle = DSHASH_HANDLE_INVALID;
-	LWLockInitialize(&memtable->apply_lock, TP_TRANCHE_CACHE_APPLY_LOCK);
-	LWLockInitialize(&memtable->lock, TP_TRANCHE_CACHE_LOCK);
+	LWLockInitialize(
+			&memtable->apply_lock, tp_tranche_id(TP_TRANCHE_CACHE_APPLY_LOCK));
+	LWLockInitialize(&memtable->lock, tp_tranche_id(TP_TRANCHE_CACHE_LOCK));
 	memtable->cursor_gen_spill_count = 0;
 	memtable->cursor_next_blkno		 = InvalidBlockNumber;
 	memtable->cursor_next_off		 = 0;
```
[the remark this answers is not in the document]
tjgreen42: (copilot) Factored into `tp_memtable_init(TpMemtable *)` in `src/index/state.c`; all three sites (`tp_create_shared_index_state`, `tp_create_build_index_state`, `tp_finalize_build_mode`) now allocate and call it, and the "dsa_allocate does not zero memory" rationale lives on the helper. Commit 6ebbdd2.

PR #461, opened 2026-08-20
test: make reclaim-dependent assertions horizon-independent

tjgreen42:
Several regression tests asserted document or page counts taken after a `DELETE`, `VACUUM`, or `REINDEX`. Those counts are only stable when the cluster's reclaim horizon has advanced past the deleting transaction.

`GetOldestNonRemovableTransactionId()` is held back by anything `ComputeXidHorizons()` folds into `data_oldest_nonremovable`: an xmin-holding backend in our database, a backend flagged `PROC_AFFECTS_ALL_HORIZONS` (hot standby feedback), `procArray->replication_slot_xmin`, or a prepared transaction. While it is held back, deleted tuples stay `HEAPTUPLE_RECENTLY_DEAD`: VACUUM must retain them, and an index build **must still index them** to preserve MVCC for pre-existing snapshots — `heapam_handler.c` says so explicitly, and `_bt_build_callback()` spools them for the same reason.

So the affected assertions were failing against a correct extension. Confirmed on stock PostgreSQL 18.1: an equivalent **core btree** partial index reports the same post-REINDEX `reltuples` that bm25 does.

## Approach

Rather than teaching each assertion to skip itself when the horizon is pinned, most of them are rewritten so the horizon is irrelevant.

A single `VACUUM`, `VACUUM FULL`, or `REINDEX TABLE` processes every index on the table under **one** reclaim horizon. That makes a core btree over the same rows an exact oracle. Measured on PostgreSQL 18.1, unpinned vs pinned:

| Operation | Unpinned | Pinned |
|---|---|---|
| `VACUUM` | bm25 50, btree 50 | bm25 100, btree 100 |
| `VACUUM FULL` | docs 2, pkey 2 | docs 4, pkey 4 |
| `REINDEX` | bm25 20, btree 20 | bm25 30, btree 30 |

The absolute counts move; the equality does not. It is also a **stronger** assertion than the magic numbers it replaces, since it pins bm25 to core's convention rather than to a hand-picked range.

`catalog_stats.sql` and `vacuum.sql` therefore contain no horizon machinery at all — no helper, no latch table, no NOTICE-count guard. `vacuum_bitmap.sql` likewise now asserts that no deleted `alpha` content resurfaces, the invariant it actually cares about, instead of counting rows returned by an `ORDER BY ... LIMIT` that does not filter.

`segment_reclaim.sql` and `memtable_reclaim.sql` keep a guard: deferred FSM reclaim is *defined* in terms of the horizon, so it cannot be asserted without reference to it, and there is no oracle. `horizon_pinned()` covers all four sources above — note a standby with `hot_standby_feedback = on` publishes its xmin through the replication slot, not `pg_stat_activity`, which is why the pre-existing guard in `memtable_reclaim.sql` never fired in the topology that needs it.

Those two files sample the pin **both** in a pre-operation latch and inline in the assertion itself, so a pin appearing or clearing mid-window is still caught, at no extra statements. The residual case — a pin that both appears and disappears inside a single VACUUM — is not observable from SQL.

## CI coverage

The guards only do anything when the horizon is actually held back, and a single-node `make installcheck` never holds it back — so a missing guard could not fail in CI. It failed later, against a streaming standby, as a confusing count mismatch in a test that looks unrelated to replication. That is how this cluster of bugs escaped.

`test/scripts/pinned_horizon.sh` closes that gap. It reuses `replication_lib.sh` to stand up a primary plus a standby running with `hot_standby_feedback = on`, holds a `REPEATABLE READ` snapshot open on the standby, and reruns the full `REGRESS` list against the primary via `pg_regress --use-existing`. It runs in the existing shell-test step on PG 17 and 18 (~1–2 min per leg).

It asserts the topology is real before trusting a pass: the slot xmin must not advance while the standby snapshot is open and must trail an xid assigned afterwards, and `horizon_pinned()` must still return true here. Without those, a change in Postgres or in the setup would leave the job green while testing nothing.

**Verified to have teeth:** reverting `catalog_stats.sql` to its pre-fix state makes exactly that test fail under this script, with a diagnostic pointing at the fix pattern.

Supporting changes: `setup_primary()` gains a `SETUP_CREATE_EXTENSION` opt-out (pg_regress needs a bare database, since the tests create and drop the extension themselves; default behaviour unchanged), and the Makefile gains `test-pinned-horizon` plus `print-regress`.

## Verification

Full suite run both ways against PostgreSQL 18.1 — normally, and against a primary whose horizon is pinned by a hot-standby-feedback standby holding an open snapshot:

| Run | Before | After |
|---|---|---|
| Baseline | 75/75 | **75/75** |
| Pinned horizon | 70/75 | **75/75** |

Before this change the pinned run failed `catalog_stats`, `vacuum`, `vacuum_bitmap`, `memtable_reclaim`, and `segment_reclaim`. Also verified via `make format-check`. No product code is changed.

## Related

Explains a cluster of Orion CI bugs whose auto-generated RCAs attributed them to extension defects: `catalog_stats` (ADO 5539172), `vacuum_bitmap` (5540066), `memtable_reclaim` (5396714), `vacuum_extended` (5499619). Notably 5539172's suggested fix — skipping `!tupleIsAlive` in `tp_build_callback` — would have dropped index entries for rows older snapshots can still legitimately read, a silent wrong-results bug.

Note `vacuum_extended` (5499619) passes under the pinned harness, so its failure has a different trigger and is not addressed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,6 +208,17 @@ jobs:
           echo "✗ WAL audit tests failed"
           exit 1
         fi
+        # Reruns the SQL suite against a primary whose reclaim horizon is
+        # pinned by a hot-standby-feedback standby. Catches assertions on
+        # counts taken after DELETE/VACUUM/REINDEX that a single-node
+        # installcheck cannot, since it never pins the horizon.
+        echo "Running pinned-horizon regression tests..."
+        if (cd test/scripts && ./pinned_horizon.sh); then
+          echo "✓ Pinned-horizon tests passed"
+        else
+          echo "✗ Pinned-horizon tests failed"
+          exit 1
+        fi
 
         # hypertable.sh needs a PostgreSQL instance with TimescaleDB preloaded
         echo "Running hypertable tests..."
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,16 @@ test-reindex:
 	@echo "Running multi-backend reindex regression tests (issue #390)..."
 	@cd test/scripts && ./multi_backend_reindex.sh
 
-test-shell: test-concurrency test-recovery test-segment test-cic test-multi-index test-reindex
+test-pinned-horizon:
+	@echo "Running pinned-horizon regression tests..."
+	@cd test/scripts && ./pinned_horizon.sh
+
+# Print the REGRESS list so shell tests can drive pg_regress without
+# duplicating it.
+print-regress:
+	@echo $(REGRESS)
+
+test-shell: test-concurrency test-recovery test-segment test-cic test-multi-index test-reindex test-pinned-horizon
 	@echo "All shell-based tests completed"
 
 test-all: test test-shell
@@ -377,4 +386,4 @@ help:
 	@echo "  make test-all"
 	@echo "  make format"
 
-.PHONY: test clean-test-dirs installcheck test-concurrency test-recovery test-segment test-stress test-cic test-chinese test-replication test-replication-extended test-logical-replication test-multi-index test-reindex test-shell test-all expected lint-format format format-check format-diff format-single coverage coverage-build coverage-clean coverage-report help
+.PHONY: test clean-test-dirs installcheck test-concurrency test-recovery test-segment test-stress test-cic test-chinese test-replication test-replication-extended test-logical-replication test-multi-index test-reindex test-pinned-horizon print-regress test-shell test-all expected lint-format format format-check format-diff format-single coverage coverage-build coverage-clean coverage-report help
--- a/test/expected/catalog_stats.out
+++ b/test/expected/catalog_stats.out
@@ -1,21 +1,6 @@
 -- Test that BM25 indexes report correct catalog statistics
 -- (pg_class.reltuples, pg_class.relpages, pg_stat_user_indexes)
 CREATE EXTENSION IF NOT EXISTS pg_textsearch;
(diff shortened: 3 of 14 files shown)


tjgreen42:
Closing this. The added test machinery isn't worth its complexity relative to the problem it solves.

The Orion/HorizonDB failures are being handled at the fork level instead, by dropping the affected tests from `REGRESS` (ADO PR 2263310), which is the same treatment already applied to `memtable_reclaim`, `segment_reclaim` and `explicit_index`.

Worth preserving from this work, independent of the patch:

- The reported failures are **not a defect**. Non-concurrent `REINDEX`/index build scans the heap with `SnapshotAny` and classifies deleted tuples via `HeapTupleSatisfiesVacuum()`. Above the reclaim horizon a deleted tuple is `HEAPTUPLE_RECENTLY_DEAD` and must still be indexed for MVCC correctness — core nbtree's `_bt_build_callback` behaves identically. Any change that skips non-alive tuples during the build would cause silent wrong results.
- Only the tests' hardcoded doc counts are horizon-dependent. Reproducible on stock PostgreSQL with a single idle `REPEATABLE READ` snapshot; no Orion required.
- A standby with `hot_standby_feedback = on` publishes its xmin via `pg_replication_slots.xmin`, **not** `pg_stat_activity.backend_xmin`. That's why the guard added in #449 never fired on Orion.

Branch `test/horizon-independent-reclaim-assertions` is left in place if any of it is wanted later.

PR #462, opened 2026-08-21
perf: bound VACUUM CTID reads

### LINKED ISSUE #320: perf: VACUUM identify phase should avoid bulk CTID preload
## Summary

`tp_vacuum_identify_affected` opens each segment with `load_ctids=true`, which bulk-loads the entire CTID arrays into memory (6 bytes/doc). For a 1M-doc segment that's ~6MB. This is the same pattern the scoring path avoids by deferring CTID resolution.

## Current behavior

The identify phase iterates every doc_id, calls `tp_segment_lookup_ctid` per doc to get the CTID, then passes it to the VACUUM callback. With `load_ctids=true`, this hits the cached arrays (fast lookup but large allocation).

## Proposed optimization

Open segments with `load_ctids=false`, then read the CTID arrays page-by-page through the segment reader in a single sequential pass. Process docs in batches as each page is read. This avoids both the bulk allocation and per-doc random I/O.

## Measuring impact

Vacuum benchmarks (in progress) will provide the baseline to measure this against.

the-osiris:
## Summary

- open VACUUM segments without preloading their complete CTID mapping
- read CTID block numbers and tuple offsets in bounded sequential batches
- reuse the same scratch buffers across batches and segment boundaries
- add multi-page, multi-segment VACUUM regression coverage

Fixes #320.

## Why

The VACUUM identify phase previously called `tp_segment_open_ex(..., true)`.
That loaded two arrays for every document in a segment:

- 4 bytes for the PostgreSQL block number
- 2 bytes for the tuple offset

The allocation therefore grew by roughly six bytes per document. An
8.84-million-document segment needed about 50.6 MiB, while larger segments
could require hundreds of megabytes or several gigabytes.

VACUUM consumes these CTIDs sequentially, so retaining the complete mapping is
unnecessary. This change keeps the working set bounded independently of segment
size while preserving document order and the existing cleanup callbacks.

## Design

`tp_vacuum_identify_affected()` now opens each segment with
`load_ctids=false`. It reads page numbers and tuple offsets through the existing
segment storage primitives in fixed-size batches, processes the resulting CTIDs
in order, and reuses the buffers for the next batch and segment.

The change does not add a public API, dependency, configuration option, generic
iterator, or on-disk format change.

## Validation

- all 75 SQL regression tests passed
- the full concurrency suite passed
- formatting passed
- a five-run MSMARCO A/B benchmark produced identical final row counts, index
  sizes, and segment layouts in all ten runs, with no errors

| VACUUM scenario | Baseline median | Patched median | Delta |
| --- | ---: | ---: | ---: |
| Concentrated delete | 5.727 s | 5.643 s | -1.45% |
| Uniform delete | 8.391 s | 7.447 s | -11.25% |
| Uniform update | 7.260 s | 7.072 s | -2.59% |
| **Total** | **20.871 s** | **20.156 s** | **-3.42%** |

The timing ranges overlap, so these results show no measurable regression
rather than a demonstrated speedup. Process-level RSS sampling could not
reliably isolate the short-lived CTID allocation from PostgreSQL shared mappings
and allocator behavior; the bounded-memory property follows directly from
replacing the proportional arrays with fixed reusable buffers.

### DIFF
--- a/test/sql/vacuum_bitmap.sql
+++ b/test/sql/vacuum_bitmap.sql
@@ -347,4 +347,31 @@ SELECT bm25_dump_index('vb_dump_idx') LIKE '%Alive: 10 / 10 docs%'
 
 DROP TABLE vb_dump;
 
+-- =============================================================
+-- Test 11: Batched CTID reads across pages and segments
+-- =============================================================
+CREATE TABLE vb_batched (id serial PRIMARY KEY, content text);
+
+INSERT INTO vb_batched (content)
+SELECT 'bounded batch ' || i
+FROM generate_series(1, 3000) i;
+
+CREATE INDEX vb_batched_idx ON vb_batched
+    USING bm25 (content) WITH (text_config = 'english');
+
+INSERT INTO vb_batched (content)
+SELECT 'bounded batch ' || i
+FROM generate_series(3001, 6000) i;
+
+DELETE FROM vb_batched WHERE id % 3 = 0;
+VACUUM vb_batched;
+
+SELECT count(*) FROM (
+    SELECT id FROM vb_batched
+    ORDER BY content <@> to_bm25query('bounded', 'vb_batched_idx')
+    LIMIT 6000
+) q;
+
+DROP TABLE vb_batched;
+
 DROP EXTENSION pg_textsearch;
--- a/test/sql/vacuum_bitmap.sql
+++ b/test/sql/vacuum_bitmap.sql
@@ -363,6 +363,8 @@ INSERT INTO vb_batched (content)
 SELECT 'bounded batch ' || i
 FROM generate_series(3001, 6000) i;
 
+SELECT bm25_spill_index('vb_batched_idx');
+
 DELETE FROM vb_batched WHERE id % 3 = 0;
 VACUUM vb_batched;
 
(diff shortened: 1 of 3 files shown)


test/sql/vacuum_bitmap.sql
```diff
@@ -347,4 +347,31 @@ SELECT bm25_dump_index('vb_dump_idx') LIKE '%Alive: 10 / 10 docs%'
...
+
+CREATE INDEX vb_batched_idx ON vb_batched
+    USING bm25 (content) WITH (text_config = 'english');
+
+INSERT INTO vb_batched (content)
+SELECT 'bounded batch ' || i
+FROM generate_series(3001, 6000) i;
+
+DELETE FROM vb_batched WHERE id % 3 = 0;
+VACUUM vb_batched;
```
tjgreen42, line 369: At this point the index is only partly on disk (the other rows inserted after creation are still in the memtable).  You might want to use `bm25_spill_index` to force a spill to disk before the VACUUM operation.

tjgreen42 (approved):
Nice work!  The memory savings is quite meaningful, I'd highlight that a bit more in the PR evaluation section.  I left one comment on the new test case.  Otherwise LGTM, let's ship it.

test/sql/vacuum_bitmap.sql
the-osiris: Pushed a new commit and updated this test case.

PR #463, opened 2026-08-23
Add a Windows MSVC build job to CI

nikitatsym:
This follows #439, where [tjgreen42 requested an MSVC CI job](https://github.com/timescale/pg_textsearch/pull/439#pullrequestreview-4963637797).

- `scripts/msvc-build.ps1` is the Windows build recipe. PGXS cannot drive MSVC and nmake cannot reuse the GNU Makefile, so instead of a second hand-maintained source list the script reads `OBJS` and `DATA` straight out of the `Makefile`.
- `ci.yml` adds a `build-msvc` matrix for PostgreSQL 17 and 18 on `windows-2022`, the same VS 2022 toolchain used to develop #439; EDB binary zips are pinned by URL and SHA-256. Like `build-test-multiple-configs`, the job compiles and links only, with no tests and no artifacts.
- A `scripts/README.md` entry documents the script and the pin-refresh procedure.

The job adds two checks beyond compilation:

- *Layout:* `src/layout_check.c`, introduced in #439 and already listed in `OBJS`, checks every packed or aligned on-disk struct against the reference GCC x86-64 layout, so MSVC layout drift fails at compile time.
- *Exports:* every symbol bound through `MODULE_PATHNAME` in a `DATA` SQL file must appear in the DLL export table. On Windows, a missing `PGDLLEXPORT` would otherwise surface only when the relevant SQL runs during extension creation or upgrade. Symbol names are compared case-sensitively, and the scan fails closed on any `MODULE_PATHNAME` occurrence it cannot parse as an explicit symbol reference.

`/W3 /WX` mirrors the Unix `-Werror`. The only suppressions are three conversion warnings (C4244, C4267, C4305) that `-Wextra` does not enable; any other warning fails the build.

Verified locally on Windows against the PostgreSQL 17 and 18 EDB trees: both build clean under `/WX` and pass the layout and export gates. [CI run 32907936088](https://github.com/nikitatsym/pg_textsearch/actions/runs/32907936088) passed all jobs, including both MSVC legs.

`build-msvc` sits in `all-tests-passed.needs`, gating merges like the other build jobs.

Out of scope: running the test suite on Windows (needs an installed tree and a running server) and official Windows binaries or packaging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,7 +314,7 @@ jobs:
         for ($attempt = 1; $attempt -le 2; $attempt++) {
           if (Test-Path -LiteralPath $zip) { Remove-Item -LiteralPath $zip -Force }
           try {
-            Invoke-WebRequest -Uri '${{ matrix.url }}' -OutFile $zip -UseBasicParsing
+            Invoke-WebRequest -Uri '${{ matrix.url }}' -OutFile $zip -UseBasicParsing -TimeoutSec 300
             $actual = (Get-FileHash -LiteralPath $zip -Algorithm SHA256).Hash.ToLowerInvariant()
             if ($actual -eq $expected) { $verified = $true; break }
             Write-Host "attempt ${attempt}: SHA256 is $actual, expected $expected"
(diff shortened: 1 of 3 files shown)


tjgreen42 (approved):
LGTM

Issue #427, opened 2026-07-13
BM25 tombstone drain reaches corrupt tombstone pages across multiple indexes


upirr:
> Note: this is an AI-assisted bug report prepared from production diagnostics. The raw observations and SQL/log outputs are real, but the interpretation may need maintainer verification.

## Bug Description

Several BM25 indexes are repeatedly failing in the deferred-free tombstone drain path with `pg_textsearch: corrupt tombstone page ...`.

This appears separate from the memtable chain-walker issue reported in #426. The failing routine here is `tp_tombstone_drain` in `segment/tombstone.c`; the invariant appears to be that every block reachable from `metap->pending_free_head` / `TpTombstonePageData.next_page` must still be a valid tombstone page.

Observed failures are persistent per index/page and occur across multiple BM25 child indexes.

## Environment

- **PostgreSQL version**: PostgreSQL 18.4 (Debian 18.4-1.pgdg13+1) on aarch64-unknown-linux-gnu
- **pg_textsearch version**: 1.3.1 (`pg_extension.extversion = 1.3.1`, `pg_textsearch.library_version = 1.3.1`)
- **Operating system**: Debian-based Linux / aarch64

## Steps to Reproduce

No minimal reproduction yet. Observed in production on partitioned BM25 indexes during normal write/update processing.

1. Maintain large BM25 child indexes on text/content chunk tables.
2. Run normal write paths that replace chunk rows / update embeddings.
3. The write path triggers tombstone drain and fails with the errors below.

Representative affected indexes/pages over a 24h log window:

    idx_8be1499d410bd40f_content_bm25: corrupt tombstone page 2167
    idx_a180d645af545c48_content_bm25: corrupt tombstone page 1351
    idx_cb0efb9afb809963_content_bm25: corrupt tombstone page 2737
    idx_79e7c4838b1ebb2d_content_bm25: corrupt tombstone page 1106
    idx_4214310855c19b6f_content_bm25: corrupt tombstone page 944

We observed about 89 primary DB errors matching this pattern in 24 hours.

Some affected index/table sizes:

    idx_8be1499d410bd40f_content_bm25
      table: chunks_contracts_p_shared_7
      index size: ~1951 MB
      table total size: ~37 GB

    idx_a180d645af545c48_content_bm25
      table: fts_chunks_contracts_p_shared_9
      index size: ~996 MB
      table total size: ~4315 MB

    idx_cb0efb9afb809963_content_bm25
[...]

Issue #380, opened 2026-05-15
Standby-safe horizon for displaced segment pages during merge


tjgreen42:
## Summary

When `merge.c` compacts level-N segments into a level-(N+1) segment,
the old level-N segment header pages are unlinked from the metapage
and the new segment is published in their place. Until a reclaim
mechanism (see orphan chain page issue) ships, those displaced pages
hang around on disk forever.

Even once we add the reclaim mechanism, replicas need a standby-safe
horizon: a primary backend may be ready to recycle a displaced page
because its `merged_at_xid` is past `RecentGlobalXmin` on the primary,
but a long-running query on a hot standby may still be reading from
the *old* segment via the same physical pages. Recycling them under
that scan corrupts the standby's results.

Predates this PR; the memtable-v2 work just made it more visible
because the orphan chain page issue uses the exact same machinery.

## Proposed approach

Adopt PG's `XLOG_BTREE_REUSE_PAGE` pattern: emit a tiny WAL record
on the primary at recycle time carrying the page's `latestRemovedXid`,
and have the standby's recovery conflict resolution defer recycling
until that XID is past all snapshots on the standby.

## Acceptance criteria

- A long-running standby query against a recently-merged segment
  does not observe partially-overwritten pages.
- Primary recycles displaced segment pages once both `RecentGlobalXmin`
  on the primary and all standby snapshots are past the page's
  `merged_at_xid`.

PR #465, opened 2026-08-24
fix: hold index lock while freeing drained tombstone blocks

tjgreen42:
## Problem

The `Sanitizer test` workflow failed on `main` (run [32755889405](https://github.com/timescale/pg_textsearch/actions/runs/32755889405), both PG17.2 and PG18.1) in `test/scripts/vacuum_concurrent_merge.sh`:

```
ERROR:  could not read blocks 296..296 in file "base/16384/16450": read only 0 of 8192 bytes
CONTEXT:  while cleaning up index "docs_bm25" of relation "public.docs"
```

No ASAN/UBSAN report accompanied it — this is a plain read-past-EOF.

## Root cause

`tp_tombstone_drain()` unlinks a past-horizon tombstone page from the deferred-free chain (#380), **releases the per-index LWLock**, and only then frees the blocks that tombstone parked.

Once unlinked, those blocks are invisible to `tp_tombstone_max_used_block()`, so a concurrent `bm25_force_merge` → `tp_truncate_dead_pages()` → `RelationTruncate()` — which runs under that same lock — truncates the relation below them. The subsequent `ReadBuffer()` in `tp_record_free_index_page()` is then past EOF.

The error is the *benign* outcome. If another backend extended the relation after the truncate, the unlocked free would stamp `TP_FREE_PAGE_MAGIC` on a block already handed to a live structure and push it into the FSM. That stamp is WAL-logged, so the corruption would replicate to standbys.

### Where it came from

Not #380, and **not a regression from #462**. `f69d22d7` (#380) introduced the release-before-free, but the free was then a pure `RecordFreeIndexPage()` — an FSM push that never touches the page, so running it unlocked was harmless. `08048d29` (#429, #427) swapped in the buffer-reading `tp_record_free_index_page()` and kept the early release, which is what made the window exploitable.

A/B against #462's parent (`c3bb8e2c`) reproduces the failure identically (1/8 runs); #462 only shifted VACUUM timing.

### Evidence

Built a clang+ASAN PostgreSQL 18 matching the sanitizer workflow and reproduced the CI error locally (~1 failure per 5 runs; never reproduces on a non-sanitizer build). Backtrace via `backtrace_functions`:

```
md_readv_report <- mdreadv <- ReadBuffer
  <- tp_record_free_index_page  (src/index/freepage.c:39)
  <- tp_tombstone_drain         (src/segment/tombstone.c)
  <- tp_vacuumcleanup           (src/access/vacuum.c:1229)
```

Temporary instrumentation in the drain caught the relation shrinking mid-drain, immediately before the error:

```
TPDBG drain listed blk=138 victim=110 nblocks_entry=149 nblocks_now=114
ERROR:  could not read blocks 138..138 ... read only 0 of 8192 bytes
```

## Fix

Hold the per-index LWLock EXCLUSIVE from the unlink through the victim's page frees, making the drain and `tp_truncate_dead_pages()` mutually exclusive. That mutual exclusion is the whole fix; the rest is hardening.

The extra hold is bounded by one tombstone page's block list (`TP_TOMBSTONE_CAPACITY`, ~2036 blocks). Freeing *before* the unlink would shorten it but is not safe: a still-chained tombstone's blocks could be claimed from the FSM and then freed a second time by a later drain of that same tombstone.

The free loop deliberately has no cancellation point — the tombstone is already unlinked, so erroring out part-way would strand the remaining blocks with nothing left to free them. The existing `CHECK_FOR_INTERRUPTS` at the head of the drain loop cancels between tombstones, where the chain is consistent.

`Assert(!own_lock || LWLockHeldByMeInMode(...))` pins the invariant, so a future change that moves the release back up trips deterministically in any `--enable-cassert` build instead of waiting for the timing window.

Also re-read `RelationGetNumberOfBlocks()` once per drain iteration rather than caching it in a C local for the whole call. To be clear about what this does and does not do: it is **not** what makes the frees safe — a fresh bound still goes stale between the check and the free, and the lock is the only real guard. It just keeps the defensive out-of-range check honest across the gap where the lock is dropped between iterations.

## Testing

- `make installcheck`: 75/75 pass (assert-enabled PG18).
- `make format-check`: clean.
- ASAN `vacuum_concurrent_merge.sh`: 60 consecutive runs pass, against a baseline of ~3 failures per 27 runs. Useful, but a clean run of a timing race is weak evidence on its own — the real argument is the mutual-exclusion invariant, plus the assert that now enforces it deterministically.

## CI coverage

The gap that let this reach `main` was the *combination*, not sanitizers alone:

- `ci.yml` runs `vacuum_concurrent_merge.sh` in the plain, non-sanitizer job — the window there is too narrow to hit (10/10 pass locally on a normal build).
- `ci.yml`'s own sanitizer job ran the regression tests plus `concurrency.sh` and `concurrent_duplicate_read.sh`, but not `vacuum_concurrent_merge.sh`.
- Only `sanitizer-build-and-test.yml`, which is `main`-only, ran it *under* sanitizers.

This PR closes that:

- Run `vacuum_concurrent_merge.sh` in `ci.yml`'s sanitizer job, at full scale. The job has ample budget — it finishes in ~14m against a 45m timeout. (A first attempt scaled it to 0.35 and completed in 16s, which verified the script *runs* rather than that the race is absent; not useful for a timing race. A `TEST_SIZE_MULTIPLIER` knob remains, for constrained runners.) That job builds with `--enable-cassert`, so the new assert is live there too.
- Grep for the past-EOF signature explicitly. The race can surface in autovacuum, where it is server-log only and would otherwise slip past the client exit-code check.
- Preserve the server and client logs on failure. The script's cleanup trap deleted its data dir, so the original failing run's CI artifacts contained no server log at all.
- Upload `test/tmp_*` as failure artifacts. Every shell test writes there, but all three workflows only globbed `test/scripts/tmp_*`, so shell-test logs were never actually collected.
- Ignore test scratch dirs and PGXS `.deps/` output, neither of which was ignored.

## Follow-ups (not in this PR)

- #468 — `tp_record_free_index_page()` reads the block with no bounds check of its own; a WARNING-and-skip there would turn any future stale-block bug into a log line instead of an `ERROR`.
- #469 — the three tombstone chain walks don't bound `cur` against `nblocks` before reading, so a corrupt `next_page` past EOF errors out before reaching the magic check that would otherwise self-heal.
- A deterministic injection test for drain-vs-truncate (the assert covers the regression case, so this is lower priority).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -357,7 +357,7 @@ jobs:
   sanitizer:
     name: PG${{ matrix.pg }} Sanitizer
     runs-on: ubuntu-22.04
-    timeout-minutes: 45
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
@@ -517,6 +517,18 @@ jobs:
           cd test/scripts && ./concurrent_duplicate_read.sh
         "
 
+        # Covers the VACUUM-vs-merge races (#411, #465).  Only this
+        # job runs it under sanitizers on a PR, where the windows are
+        # wide enough to hit; scaled down because every page mutation
+        # is slow under ASAN.
+        echo "Running VACUUM concurrency tests under sanitizer..."
+        timeout 900s bash -c "
+          export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
+          export PGHOST=$PWD/tmp_sanitizer_test
+          export PGPORT=55433
+          cd test/scripts && TEST_SIZE_MULTIPLIER=0.35 ./vacuum_concurrent_merge.sh
+        "
+
         echo "Running crash recovery tests under sanitizer..."
         timeout 600s bash -c "
           export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
@@ -614,6 +626,7 @@ jobs:
         path: |
           tmp_sanitizer_test/
           test/scripts/tmp_*
+          test/tmp_*
         retention-days: 7
 
   coverage:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -517,10 +517,9 @@ jobs:
           cd test/scripts && ./concurrent_duplicate_read.sh
         "
 
-        # Covers the VACUUM-vs-merge races (#411, #465).  Only this
-        # job runs it under sanitizers on a PR, where the timing
-        # windows are wide enough to hit.  Run at full scale: the
-        # scaled-down variant finishes too fast to exercise the race.
+        # Covers the VACUUM-vs-merge races (#411, #465).  This is the
+        # only sanitized run of it on a PR, where the timing windows
+        # are wide enough to hit.
         echo "Running VACUUM concurrency tests under sanitizer..."
         timeout 900s bash -c "
           export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -357,7 +357,7 @@ jobs:
   sanitizer:
     name: PG${{ matrix.pg }} Sanitizer
     runs-on: ubuntu-22.04
-    timeout-minutes: 60
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
@@ -518,15 +518,15 @@ jobs:
         "
 
         # Covers the VACUUM-vs-merge races (#411, #465).  Only this
-        # job runs it under sanitizers on a PR, where the windows are
-        # wide enough to hit; scaled down because every page mutation
-        # is slow under ASAN.
+        # job runs it under sanitizers on a PR, where the timing
+        # windows are wide enough to hit.  Run at full scale: the
+        # scaled-down variant finishes too fast to exercise the race.
         echo "Running VACUUM concurrency tests under sanitizer..."
         timeout 900s bash -c "
           export PATH=\"$HOME/$PG_INSTALL_DIR/bin:\$PATH\"
           export PGHOST=$PWD/tmp_sanitizer_test
           export PGPORT=55433
-          cd test/scripts && TEST_SIZE_MULTIPLIER=0.35 ./vacuum_concurrent_merge.sh
+          cd test/scripts && ./vacuum_concurrent_merge.sh
         "
 
         echo "Running crash recovery tests under sanitizer..."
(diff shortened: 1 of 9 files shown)


tjgreen42:
Addressed the review feedback in cbaa4723.

**`tombstone.h` contract (from @claude's review):** fixed. The `own_lock=true` doc still promised readers "never wait more than one unlink", which stopped being true the moment the fix extended the hold across the frees. It now states the actual bound — one chain walk, one unlink, and up to `TP_TOMBSTONE_CAPACITY` page frees.

**Self-review changes on top of that:**

- `Assert(!own_lock || LWLockHeldByMeInMode(&state->shared->lock, LW_EXCLUSIVE))` before the free loop. The fix is an invariant that is invisible in the code — nothing stopped a later change from moving `tp_release_index_lock()` back up and silently reintroducing the race. It now trips deterministically in any `--enable-cassert` build, including the sanitizer CI job, instead of waiting for the timing window. Plus the `own_lock=true => state != NULL` precondition the header documents.
- Corrected the `RelationGetNumberOfBlocks()` comment and commit message. The original wording implied the per-iteration re-read is what makes the frees safe. It is not — a fresh bound still goes stale between the check and the free, and the lock is the only real guard. (Worth recording separately: the value was never actually stale for a second reason. `smgrnblocks_cached()` declines the smgr size cache outside recovery, so `RelationGetNumberOfBlocks()` re-stats the file. Only the plain C local was stale.)
- Documented why the free loop has no cancellation point (the tombstone is already unlinked, so erroring out part-way strands the rest with nothing left to free them) and why free-before-unlink is not an alternative (a still-chained tombstone's blocks could be FSM-claimed and then freed a second time).

**CI, which is the part that actually let this reach `main`:** `ci.yml`'s sanitizer job now runs `vacuum_concurrent_merge.sh` at `TEST_SIZE_MULTIPLIER=0.35`. Previously the only *sanitized* run of that script was in the `main`-only workflow. The script also greps for the past-EOF signature explicitly (the race can surface in autovacuum, where it is server-log only and slips past the client exit-code check) and preserves logs on failure — the original failing run uploaded no server log at all, because the cleanup trap deletes the data dir. Also fixed the failure-artifact glob in all three workflows: they globbed `test/scripts/tmp_*`, but every shell test writes to `test/tmp_*`, so shell-test logs were never collected.

**Validation of the final tree:** 60/60 consecutive ASAN `vacuum_concurrent_merge.sh` runs pass (baseline ~3 failures per 27); `make installcheck` 75/75 on an assert-enabled PG18; `make format-check` clean.

Two hardening items are split out rather than scope-creeping this PR: #468 (`tp_record_free_index_page()` has no bounds check of its own) and #469 (the tombstone chain walks don't bound the next-page pointer, so out-of-range corruption errors before reaching the magic check that would self-heal).

the-osiris (approved):

PR #476, opened 2026-08-29
Add pg_durable compaction adapter

tjgreen42:
## Summary

Add an optional pg_durable adapter for per-index background compaction.
pg_textsearch's dispatch stays scheduler-neutral; the adapter supplies the
operator scripts and the two stale-safe entry points a durable scheduler
needs.

## Merge blocker

**Do not merge until [microsoft/pg_durable#354](https://github.com/microsoft/pg_durable/pull/354)
lands in a release.** The adapter needs the seven-argument `df.start` with
`on_failure => 'continue'`. No released pg_durable provides it: v0.2.7
shipped without #354, which is still an open draft. Both scripts check the
exact `df.start` signature before any managed side effect, so they refuse to
install rather than half-configure. The capability checks must gain a
released version floor once one exists.

## Design

- `01_setup_role.sql` creates and validates a least-privilege compactor role
  in a single transaction, making no change when pg_durable capabilities are
  absent or drifted.
- `02_wrapper.sql` installs a SECURITY DEFINER request wrapper with a fixed
  search path, target validation, and owner-only helper privileges.
- Each BM25 index is submitted to its own pg_durable loop, so pg_textsearch
  owns the per-index identity rather than requiring a keyed scheduling API
  upstream.
- `bm25_compact_step_if_current()` and `bm25_needs_compaction_if_current()`
  bind work to the database, index OID, tablespace and relfilenumber, and
  return false when the target is stale, so a stale job cannot compact a
  replacement relation.
- Each loop iteration calls `bm25_compact_step_if_current()` in its own
  transaction, releasing the per-index lock between merge batches.
- Install and upgrade SQL revoke helper access from nonowners, including
  grants inherited from `ALTER DEFAULT PRIVILEGES`.

## Test coverage

`test/scripts/durable_compaction.sh` runs in two modes. The static mode
checks wrapper identity, body hashes, extension ownership, privilege
boundaries, callback wiring, and package contents. It needs no pg_durable,
so it gates `installcheck` and `test-local` alongside the other source
guards. The live mode (`make test-durable`) additionally covers pre-commit
dispatch, that PREPARE dispatches nothing, precise skipping against a build
lacking the failure policy, and full execution against #354. It is excluded
from `test-all` because the CI image has no pg_durable.

## Known limitations

- Submission is not deduplicated across transactions. Every committing
  writer that spills can start another workflow for the same index; the
  per-index lock serializes the work but the instances still accumulate.
  A singleton admission key is needed before this is production-ready.
- A target that goes stale through `REINDEX` or `ALTER INDEX ... SET
  TABLESPACE` stops compacting until a later spill registers a new
  identity. There is no reconciliation sweep.
- The stale and no-work cases are both `false`, so a scheduler cannot
  distinguish them.

## Stack position

Fourth layer of stack #479, on top of #477's dispatch. Draft pending the
upstream release. The final layer adds the singleton recovery backstop and
operational packaging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,8 @@ REGRESS = abort aerodocs basic binary_io bmw bmw_skip_advance bulk_load cache_ap
 REGRESS_OPTS = --inputdir=test --outputdir=test
 
 PG_CONFIG ?= pg_config
-PGXS := $(shell $(PG_CONFIG) --pgxs)
+PG_CONFIG_PATH := $(realpath $(shell command -v $(PG_CONFIG) 2>/dev/null))
+PGXS := $(shell $(PG_CONFIG_PATH) --pgxs)
 include $(PGXS)
 
 # SQL regression tests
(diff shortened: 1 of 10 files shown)


tjgreen42:
Superseded by #478, which provides the scheduler-neutral periodic compaction sweep.

PR #480, opened 2026-09-01
Add Boolean tsquery index scans

### LINKED ISSUE #315: Boolean queries (unranked AND/OR/NOT)
## Summary

Support unranked boolean query semantics using AND/OR/NOT operators, leveraging Postgres's existing `to_tsquery` infrastructure.

## What's needed

- A new query evaluation mode that returns yes/no matches instead of BM25 scores
- Parse boolean operators from `tsquery` input (AND, OR, NOT, phrases via `<->`)
- No index format changes required — this is purely a query processing addition

## Current state

- No requests for this yet
- Workaround available: post-filtering with `@@` operator after BM25 retrieval, or using GIN indexes with `tsvector`/`tsquery` directly

## Use cases

- Exact match filtering ("must contain X AND Y")
- Exclusion queries ("contains X but NOT Y")
- Combining boolean filters with BM25 ranking in a single index scan

## Recommendation

Wait — no demand yet and the existing Postgres `tsvector`/GIN infrastructure handles this use case.

the-osiris:
## Summary

- add `text @@ tsquery` scans to the BM25 operator family
- evaluate PostgreSQL Boolean query trees against existing postings
- support AND, OR, NOT, pure-negative queries, parameters, and conservative heap rechecks for phrases, prefixes, and weights
- preserve zero-lexeme documents through build, spill, merge, and VACUUM paths
- register the operator for fresh installs and 1.4-to-1.5 upgrades

## Design

This reuses PostgreSQL's `TSQuery` parser and ternary evaluator rather than introducing another query language or changing the posting-list format. Full-universe scans stream results through a temporary `BufFile` so result storage remains bounded.

Boolean filtering and BM25 ranking in one index scan are intentionally out of scope.

## Validation

- extension build succeeds
- focused `boolean_queries` regression passes with the worktree module
- isolated 1 MB zero-lexeme build-memory regression passes
- upgrade catalog assertion passes on PostgreSQL 17

## Known blockers

- the existing unbounded BM25 ranked query in `parallel_build` reproducibly crashes the backend and must be fixed before this PR is ready for review
- the text-search configuration compatibility check needs correction because `tsquery` does not retain its source configuration
- high-frequency posting membership sets are currently unbounded; a bounded evaluation strategy is required before merge
- focused performance benchmarks remain to be run after correctness blockers are resolved

Closes timescale/pg_textsearch#315

### DIFF
--- a/sql/pg_textsearch--1.4.0--1.5.0-dev.sql
+++ b/sql/pg_textsearch--1.4.0--1.5.0-dev.sql
@@ -61,3 +61,6 @@ $$;
 
 COMMENT ON FUNCTION @extschema@.bm25_needs_compaction(regclass) IS
     'Report whether any level holds at least segments_per_level segments. Advisory only: a level whose segments are all over budget is reported as full even though bm25_compact_step has no way to reduce it, so this must not be used on its own as a retry condition.';
+
+ALTER OPERATOR FAMILY @extschema@.text_bm25_ops USING bm25
+    ADD OPERATOR 1 pg_catalog.@@ (text, pg_catalog.tsquery);
--- a/src/access/boolean.c
+++ b/src/access/boolean.c
@@ -21,6 +21,8 @@
 #include "segment/alive_bitset.h"
 #include "segment/io.h"
 
+#define TP_BOOLEAN_MAX_PREFIX_ITERATORS 256
+
 typedef struct TpBooleanTerm
 {
 	char *lexeme;
@@ -42,7 +44,15 @@ typedef enum TpBooleanCandidateKind
 	TP_BOOLEAN_CANDIDATE_ALL,
 	TP_BOOLEAN_CANDIDATE_TERM,
 	TP_BOOLEAN_CANDIDATE_UNION,
+	TP_BOOLEAN_CANDIDATE_PREFIX,
+	TP_BOOLEAN_CANDIDATE_PREFIX_BITMAP,
 } TpBooleanCandidateKind;
+
+typedef struct TpBooleanPrefixHeapNode
+{
+	uint32 iterator_index;
+	uint32 doc_id;
+} TpBooleanPrefixHeapNode;
 
 typedef struct TpBooleanCandidateStream
 {
@@ -74,6 +84,22 @@ typedef struct TpBooleanCandidateStream
 			bool							 left_valid;
 			bool							 right_valid;
 		} union_stream;
+
+		struct
+		{
+			TpSegmentPostingIterator *iterators;
+			TpBooleanPrefixHeapNode	 *heap;
+			uint32					  iterator_count;
+			uint32					  heap_size;
+			bool					  initialized;
+		} prefix;
+
+		struct
+		{
+			uint8 *bits;
+			uint32 next_doc_id;
+			uint32 num_docs;
+		} prefix_bitmap;
 	} state;
 } TpBooleanCandidateStream;
 
@@ -176,9 +202,9 @@ tp_boolean_extract_terms(TSQuery query)
 			TpBooleanTerm *term;
 
 			/*
-			 * Postings do not retain lexeme weights and the index cannot
-			 * enumerate dictionary prefixes yet.  Treat these operands as
-			 * unknown over an all-document candidate set and let the heap
+			 * Postings do not retain lexeme weights, and the memtable source
+			 * cannot enumerate prefixes. Treat these operands as unknown
+			 * over an all-document memtable candidate set and let the heap
 			 * recheck evaluate their exact semantics.
 			 */
 			if (operand->prefix || operand->weight != 0)
@@ -297,6 +323,136 @@ tp_boolean_create_all_stream(TpSegmentReader *reader)
 static void tp_boolean_free_candidate_stream(TpBooleanCandidateStream *stream);
 
 static TpBooleanCandidateStream *
+tp_boolean_create_union_stream(
+		TpSegmentReader			 *reader,
+		TpBooleanCandidateStream *left,
+		TpBooleanCandidateStream *right)
+{
+	TpBooleanCandidateStream *stream = palloc0(sizeof(*stream));
+
+	stream->kind = TP_BOOLEAN_CANDIDATE_UNION;
+	stream->estimate =
+			Min((uint64)reader->header->num_docs,
+				left->estimate + right->estimate);
+	stream->state.union_stream.left	 = left;
+	stream->state.union_stream.right = right;
+	return stream;
+}
+
+static bool
+tp_boolean_prefix_heap_less(
+		const TpBooleanPrefixHeapNode *left,
+		const TpBooleanPrefixHeapNode *right)
+{
+	return left->doc_id < right->doc_id ||
+		   (left->doc_id == right->doc_id &&
+			left->iterator_index < right->iterator_index);
+}
+
+static void
+tp_boolean_prefix_heap_sift_down(
+		TpBooleanPrefixHeapNode *heap, uint32 heap_size, uint32 parent)
+{
+	TpBooleanPrefixHeapNode node = heap[parent];
+
+	while (parent < heap_size / 2)
+	{
+		uint32 child = parent * 2 + 1;
+
+		if (child + 1 < heap_size &&
+			tp_boolean_prefix_heap_less(&heap[child + 1], &heap[child]))
+			child++;
+		if (!tp_boolean_prefix_heap_less(&heap[child], &node))
+			break;
+		heap[parent] = heap[child];
+		parent		 = child;
+	}
+
+	heap[parent] = node;
+}
+
+static void
+tp_boolean_prefix_stream_init(TpBooleanCandidateStream *stream)
+{
+	for (uint32 i = 0; i < stream->state.prefix.iterator_count; i++)
+	{
+		TpSegmentPosting *posting;
+
+		if (tp_segment_posting_iterator_next(
+					&stream->state.prefix.iterators[i], &posting))
+			stream->state.prefix.heap[stream->state.prefix.heap_size++] =
+					(TpBooleanPrefixHeapNode){
+							.iterator_index = i,
+							.doc_id			= posting->doc_id,
+					};
+	}
+
+	for (uint32 i = stream->state.prefix.heap_size / 2; i > 0; i--)
+		tp_boolean_prefix_heap_sift_down(
+				stream->state.prefix.heap,
+				stream->state.prefix.heap_size,
+				i - 1);
+
+	stream->state.prefix.initialized = true;
+}
+
+static void
+tp_boolean_prefix_stream_advance_root(TpBooleanCandidateStream *stream)
+{
+	TpBooleanPrefixHeapNode *heap = stream->state.prefix.heap;
+	TpSegmentPosting		*posting;
+	uint32					 iterator_index = heap[0].iterator_index;
+
+	if (tp_segment_posting_iterator_next(
+				&stream->state.prefix.iterators[iterator_index], &posting))
+		heap[0].doc_id = posting->doc_id;
+	else
+		heap[0] = heap[--stream->state.prefix.heap_size];
+
+	if (stream->state.prefix.heap_size > 0)
+		tp_boolean_prefix_heap_sift_down(
+				heap, stream->state.prefix.heap_size, 0);
+}
+
+static TpBooleanCandidateStream *
+tp_boolean_create_prefix_stream(
+		TpSegmentReader *reader, const char *prefix, int prefix_length)
+{
+	TpSegmentPrefixCandidates candidates;
+	TpBooleanCandidateStream *stream = palloc0(sizeof(*stream));
+
+	tp_segment_prefix_candidates_init(
+			reader,
+			prefix,
+			prefix_length,
+			TP_BOOLEAN_MAX_PREFIX_ITERATORS,
+			&candidates);
+	stream->estimate = candidates.estimate;
+
+	if (candidates.doc_bitmap != NULL)
+	{
+		stream->kind					 = TP_BOOLEAN_CANDIDATE_PREFIX_BITMAP;
+		stream->state.prefix_bitmap.bits = candidates.doc_bitmap;
+		stream->state.prefix_bitmap.num_docs = reader->header->num_docs;
+		return stream;
+	}
+
+	if (candidates.iterator_count == 0)
+	{
+		stream->kind = TP_BOOLEAN_CANDIDATE_TERM;
+		return stream;
+	}
+
+	stream->kind						= TP_BOOLEAN_CANDIDATE_PREFIX;
+	stream->estimate					= candidates.estimate;
+	stream->state.prefix.iterators		= candidates.iterators;
+	stream->state.prefix.iterator_count = candidates.iterator_count;
+	stream->state.prefix.heap =
+			palloc_array(TpBooleanPrefixHeapNode, candidates.iterator_count);
+	return stream;
+}
+
+static TpBooleanCandidateStream *
 tp_boolean_create_candidate_stream(
 		TpBooleanEvalState *state, TpSegmentReader *reader, QueryItem *item)
 {
@@ -307,7 +463,11 @@ tp_boolean_create_candidate_stream(
 		QueryOperand *operand = &item->qoperand;
 		const char	 *lexeme  = GETOPERAND(state->query) + operand->distance;
 
-		if (operand->prefix || operand->weight != 0)
+		if (operand->prefix)
+			return tp_boolean_create_prefix_stream(
+					reader, lexeme, operand->length);
+
+		if (operand->weight != 0)
 			return tp_boolean_create_all_stream(reader);
 
 		stream						   = palloc0(sizeof(*stream));
@@ -353,14 +513,7 @@ tp_boolean_create_candidate_stream(
 					(errcode(ERRCODE_DATA_CORRUPTED),
 					 errmsg("invalid operator in BM25 Boolean query")));
 
-		stream		 = palloc0(sizeof(*stream));
-		stream->kind = TP_BOOLEAN_CANDIDATE_UNION;
-		stream->estimate =
-				Min((uint64)reader->header->num_docs,
-					left->estimate + right->estimate);
-		stream->state.union_stream.left	 = left;
-		stream->state.union_stream.right = right;
-		return stream;
+		return tp_boolean_create_union_stream(reader, left, right);
 	}
 }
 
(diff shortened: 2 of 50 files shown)


the-osiris:
@claude review

src/planner/cost.c
```diff
@@ -141,6 +205,16 @@ tp_costestimate(
...
 			metap = tp_get_metapage(index_rel);
+			if (has_boolean &&
+				metap->text_config_oid != getTSCurrentConfig(true))
+			{
+				pfree(metap);
+				index_close(index_rel, AccessShareLock);
+				*indexStartupCost = get_float8_infinity();
+				*indexTotalCost	  = get_float8_infinity();
+				return;
+			}
```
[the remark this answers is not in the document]
the-osiris: This gate intentionally compares the index configuration with `default_text_search_config`, because that is the configuration PostgreSQL uses for the left-hand `text` in `text @@ tsquery`. The `tsquery` source configuration is not the compatibility boundary. Returning infinite cost here makes PostgreSQL use the sequential implementation when the index cannot preserve those text-parsing semantics; the executor repeats the check for cached/generic plans.

src/access/boolean.c
```diff
@@ -0,0 +1,920 @@
...
+						"configuration \"%s\"",
+						RelationGetRelationName(index),
+						current_config_name),
+				 errdetail(
+						 "The index uses text search configuration \"%s\".",
+						 index_config_name),
+				 errhint("Set default_text_search_config to match the index "
+						 "or use a different index.")));
+	}
+}
```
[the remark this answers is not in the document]
the-osiris: The compatibility check is for the left-hand `text`, not the already-normalized `tsquery`. PostgreSQL evaluates `text @@ tsquery` by parsing the text with `default_text_search_config`; therefore an index built with another configuration cannot safely answer it. An explicitly configured `tsquery` may differ because only its resulting lexemes participate in matching. The planner gate prevents choosing an incompatible index, and the executor check protects cached/generic plans where the session default may have changed.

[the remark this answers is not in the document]
the-osiris: The hard check is required for correctness. It validates the configuration PostgreSQL uses to parse the left-hand `text`, not the source configuration of the normalized `tsquery`. If the session default differs from the index configuration, evaluating directly from index postings can disagree with PostgreSQL `text @@ tsquery` semantics. Keeping this executor-time check also protects prepared and generic plans after the GUC changes.

src/access/handler.c
```diff
@@ -16,6 +16,7 @@
 
 #include "access/am.h"
 #include "planner/cost.h"
+#include "tsearch/ts_utils.h"
```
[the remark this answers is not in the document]
the-osiris: Good catch. I have switched this PostgreSQL header to the angle-bracket include style used by the rest of the codebase; it will be included in the next push.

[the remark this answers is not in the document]
the-osiris: Follow-up: I reverted the angle-bracket change after CI exposed an include-order hazard. clang-format moves `<tsearch/ts_utils.h>` before the project `"access/am.h"`; because the project has `src/access/am.h`, transitive PostgreSQL `<access/am.h>` resolution can then be shadowed and break the `IndexAmRoutine` ABI. Keeping this include quoted preserves the required order and passes the repository formatter.

the-osiris:
@claude review

sql/pg_textsearch--1.4.0--1.5.0-dev.sql
```diff
@@ -61,3 +61,6 @@ $$;
 
 COMMENT ON FUNCTION @extschema@.bm25_needs_compaction(regclass) IS
     'Report whether any level holds at least segments_per_level segments. Advisory only: a level whose segments are all over budget is reported as full even though bm25_compact_step has no way to reduce it, so this must not be used on its own as a retry condition.';
+
+ALTER OPERATOR FAMILY @extschema@.text_bm25_ops USING bm25
+    ADD OPERATOR 1 pg_catalog.@@ (text, pg_catalog.tsquery);
```
tjgreen42, line 66: **Copilot review — High:** This upgrade enables `@@` on existing indexes without rebuilding them. Pre-1.5 build and insert paths skipped documents with zero lexemes, so upgraded indexes have no CTIDs for historical empty or stopword-only rows. Pure-negative queries such as `!missing` will silently return false negatives. Please require `REINDEX`, or gate Boolean scans on an index capability/version until rebuilt.

src/access/scan.c
```diff
@@ -477,19 +508,57 @@ tp_gettuple(IndexScanDesc scan, ScanDirection dir)
...
+					RelationGetRelid(scan->indexRelation));
+			TpIndexMetaPage metap;
+
+			if (!index_state)
+				ereport(ERROR,
+						(errcode(ERRCODE_INTERNAL_ERROR),
+						 errmsg("could not get index state for BM25 "
+								"Boolean search")));
+
+			tp_acquire_index_lock(index_state, LW_SHARED);
```
tjgreen42, line 523: **Copilot review — High:** `tp_boolean_execute()` runs while this LWLock is held and exhausts the memtable and all segments before returning. PostgreSQL holds cancel/die interrupts for the lifetime of an LWLock, so the internal `CHECK_FOR_INTERRUPTS()` calls cannot process `statement_timeout` or client cancellation. Broad scans also block exclusive spill and compaction. Retain a stable index view under the lock, then release it before cardinality-dependent evaluation.

src/planner/cost.c
```diff
@@ -95,18 +148,29 @@ tp_costestimate(
...
-	/* Never use index without ORDER BY clause */
-	if (!path->indexorderbys || list_length(path->indexorderbys) == 0)
+	/*
+	 * Boolean filtering and ranked scans are separate execution modes.
+	 * Multiple Boolean keys and combined filtering/ranking are follow-ups.
+	 */
+	if ((!has_orderby && !has_boolean) || (has_orderby && has_boolean) ||
+		(has_boolean && list_length(path->indexclauses) != 1))
 	{
 		*indexStartupCost = get_float8_infinity();
```
tjgreen42, line 163: **Copilot review — High:** This return initializes only two of the five mandatory `amcostestimate` outputs. Core immediately consumes `indexSelectivity`, `indexCorrelation`, and `indexPages`, and computes total minus startup, so this can produce uninitialized values and `Infinity..NaN`. Infinity is also not a capability veto on PostgreSQL 18, where disabled-node ordering can still select the path. Prevent unsupported path construction, or return fully initialized valid costs.

src/access/boolean.c
```diff
@@ -0,0 +1,920 @@
...
+
+static void
+tp_boolean_add_ctid(TpBooleanTerm *term, HTAB *candidates, ItemPointer ctid)
+{
+	bool found;
+
+	(void)hash_search(term->ctids, ctid, HASH_ENTER, &found);
+	if (candidates != NULL)
+		(void)hash_search(candidates, ctid, HASH_ENTER, &found);
```
tjgreen42, line 240: **Copilot review — High:** Every query term gets a complete CTID hash, and positive queries duplicate those CTIDs into the candidate hash. A selective `rare & common` query therefore materializes the entire `common` posting set. This backend memory is not bounded by `work_mem` or `pg_textsearch.memory_limit`, and auto-spill can be disabled. This needs a bounded/spillable evaluation strategy or a smallest-conjunct candidate driver.

src/access/boolean.c
```diff
@@ -0,0 +1,920 @@
...
+
+	for (int i = 0; i < state.term_count; i++)
+		terms[i] = state.terms[i].lexeme;
+
+	memtable_source = tp_memtable_source_create_for_read(
+			index_state, scan->indexRelation, terms, state.term_count);
+	tp_boolean_collect_memtable(&state, memtable_source, candidates);
+
+	writer.query = &state;
+	writer.file	 = BufFileCreateTemp(false);
```
tjgreen42, line 853: **Copilot review — High:** All results are written to this `BufFile` and every segment is exhausted before the first tuple can be returned, but the planner reports near-zero startup cost and may derive selectivity from `LIMIT`. `LIMIT 1` and `EXISTS` can consequently choose an O(N), temp-writing path over a sequential scan that stops immediately. Stream candidates, or charge the full materialization work to startup cost.

src/access/scan.c
```diff
@@ -284,10 +301,21 @@ tp_rescan(
...
 		/* Reset scan position and state */
-		so->current_pos	 = 0;
-		so->result_count = 0;
-		so->eof_reached	 = false;
-		so->query_vector = NULL;
+		so->current_pos		= 0;
+		so->result_count	= 0;
+		so->eof_reached		= false;
+		so->query_vector	= NULL;
+		so->boolean_query	= NULL;
```
tjgreen42, line 308: **Copilot review — Medium:** PostgreSQL defines `amrescan` with `keys == NULL` as a restart using the previously supplied keys. This reset unconditionally clears `boolean_query` and `is_boolean_scan`, while reconstruction happens only when `keys` is non-NULL. A same-key restart therefore loses Boolean mode and can assert or return EOF. Preserve the copied Boolean key across NULL-key rescans.

src/planner/cost.c
```diff
@@ -141,6 +205,16 @@ tp_costestimate(
 		if (index_rel)
 		{
 			metap = tp_get_metapage(index_rel);
+			if (has_boolean &&
```
tjgreen42, line 208: **Copilot review — Medium:** Boolean path eligibility depends on `default_text_search_config`, but changing that GUC does not invalidate cached generic plans. A freshly planned statement can fall back to a sequential scan, while the cached equivalent reaches the executor-time mismatch error. The configuration check is needed for left-hand text semantics, but the plan dependency must trigger invalidation/replanning or provide a correct lossy fallback.

src/access/boolean.c
```diff
@@ -0,0 +1,920 @@
...
+			writer->query->term_count * sizeof(*cursors));
+	uint32 cache_threshold = reader->header->num_docs / 100 +
+							 (reader->header->num_docs % 100 != 0);
+	uint32 doc_id;
+
+	if (stream->estimate >= cache_threshold)
+		tp_segment_enable_ctid_lookup_cache(reader);
+
+	for (int i = 0; i < writer->query->term_count; i++)
+		cursors[i].initialized = tp_segment_posting_iterator_init(
```
tjgreen42, line 757: **Copilot review — Medium:** The scan creates one cursor per term. As those cursors seek, direct block access can retain a buffer content LWLock until the iterator advances or is freed. A sufficiently large all-matching conjunction can exceed PostgreSQL’s hard `MAX_SIMUL_LWLOCKS` limit of 200 and fail with `too many LWLocks taken`. Bound the number of simultaneous direct iterators or force copied access for multi-term evaluation.

tjgreen42 (commented):
**Copilot CLI review** — submitted through the authenticated `tjgreen42` account at the requester’s direction. I found eight actionable issues; details are inline.

sql/pg_textsearch--1.4.0--1.5.0-dev.sql
tjgreen42: @the-osiris I think we should print a WARNING rather than require a full reindex on existing indexes, since this is a small corner case, and only do so for purely negative queries.

src/access/boolean.c
tjgreen42: @the-osiris (1) Is it possible to return chunks of answers as we go, rather than the entire result set?  What does the access method interface support? (2) A simple optimization we could do for AND queries would be to sort ascending by posting list length and evaluate in that order, which would bound the result size by the length of the shortest posting list.

src/access/boolean.c
tjgreen42: @the-osiris as noted earlier, see what the options are here.  This might be a big enough work item to merit deferral to a later PR.

src/planner/cost.c
tjgreen42: The GUC in question is just for testing, I assume?  Probably can ignore this one.

src/access/boolean.c
tjgreen42: Pathological case, but some kind of (generous) cap on number of conjuncts in the query might be wise.

tjgreen42:
One other comment, the unit tests cover only the single-segment case, I believe.  Make sure you add some covering multi-segment evaluation.  See other unit tests for examples.

sql/pg_textsearch--1.4.0--1.5.0-dev.sql
the-osiris: Fixed. Pure-negative queries now warn on older indexes, and REINDEX clears the warning and restores complete results.

src/access/scan.c
the-osiris: Fixed. We now snapshot the memtable and segment roots under the shared lock, then release it before evaluating or writing results.

src/planner/cost.c
the-osiris: Fixed. Rejected paths now initialize all required cost fields and are properly disabled on PostgreSQL 18.

src/access/boolean.c
the-osiris: Fixed. Memtable candidates now use compact sorted lists, with AND driven by the smallest child and OR merged lazily. I left full result streaming for a follow-up.

src/access/boolean.c
the-osiris: Fixed. Boolean materialization is now charged to startup cost, and LIMIT no longer discounts that work. Full result streaming remains deferred.

src/access/scan.c
the-osiris: Fixed. NULL-key rescans now preserve the copied Boolean query while resetting the execution state.

src/planner/cost.c
the-osiris: Agreed, leaving this unchanged. The executor check still prevents incorrect results if a cached generic plan outlives a GUC change.

src/access/boolean.c
the-osiris: Fixed. Exact operands are capped at 64; oversized constant queries fall back to a sequential scan, while dynamic queries fail before taking index locks.

tjgreen42:
> One other comment, the unit tests cover only the single-segment case, I believe. Make sure you add some covering multi-segment evaluation. See other unit tests for examples.

@the-osiris ^^^

tjgreen42:
@the-osiris PR is looking good, but we need documentation in the README of the new functionality

tjgreen42:
Also, include in the PR description a quick summary of the performance numbers you measured vs GIN.

tjgreen42 (commented):
**Copilot CLI follow-up review** — submitted through the authenticated `tjgreen42` account at the requester’s direction.

I rechecked Kunal’s follow-up commits against the September 8 findings. Several fixes look good: unsupported paths now initialize all `amcostestimate` outputs and are disabled on PG18; Boolean materialization is charged to startup cost; copied Boolean keys survive NULL-key rescans; exact posting iterators are capped; and the SQL regression now covers multiple persisted segments plus a live memtable.

I do **not** think the PR is ready as-is, though:

1. **High — the legacy-index warning still misses mixed-negative queries.** [`tp_boolean_warn_if_incomplete()`](https://github.com/timescale/pg_textsearch/blob/7ed48b64a7d23c48523fd7c44317f0112e324720/src/access/boolean.c#L192-L201) suppresses the warning whenever the tree contains any positive-polarity operand. But `alpha | !missing` contains `alpha` and still matches empty/stopword-only documents that pre-1.5 indexes omitted. On the upgrade-test corpus it should return all three rows, while the legacy index can silently return only the indexed `alpha` row. The safe distinction already exists as `tsquery_requires_match()`: warn whenever that returns false, and add the mixed-OR upgrade regression.

2. **High — the memtable memory/lock issue is only partially fixed.** [`tp_boolean_collect_memtable_terms()`](https://github.com/timescale/pg_textsearch/blob/7ed48b64a7d23c48523fd7c44317f0112e324720/src/access/boolean.c#L355-L374) still copies every exact term’s complete posting list before the smallest candidate branch is selected. [`tp_boolean_execute()`](https://github.com/timescale/pg_textsearch/blob/7ed48b64a7d23c48523fd7c44317f0112e324720/src/access/boolean.c#L1247-L1276) performs that work, and potentially all-document collection, while holding the per-index LWLock. With `pg_textsearch.memtable_pages_threshold=0`, this remains unbounded `O(sum(doc_freq))` backend memory, blocks spill/compaction, and defers cancellation/`statement_timeout`. Prefix-only or weighted chain scans are worse: zero filtered terms causes the chain source to reconstruct the entire term dictionary even though only document CTIDs are needed. The new source guard proves the hashes disappeared, not that memory or lock duration is bounded.

3. **Medium — positive exact-term segment scans remain uncancellable.** Exact candidate/evaluator iterators use zero-copy access, and [`tp_segment_posting_iterator_load_block()`](https://github.com/timescale/pg_textsearch/blob/7ed48b64a7d23c48523fd7c44317f0112e324720/src/segment/scan.c#L405-L418) retains a buffer-content LWLock. PostgreSQL holds cancel/die interrupts until an LWLock is released, so the `CHECK_FOR_INTERRUPTS()` calls in the segment loop cannot process cancellation while these iterators remain active. The 64-operand cap fixes simultaneous lock count, not lock duration. Force copied access for Boolean exact iterators, or otherwise release direct access before interrupt checks, and add an uncompressed positive-term `statement_timeout` test.

4. **CI blocker — formatting still fails.** `make format-check` reports violations at [`src/access/scan.c:307-310`](https://github.com/timescale/pg_textsearch/blob/7ed48b64a7d23c48523fd7c44317f0112e324720/src/access/scan.c#L307-L310), matching the red `Check C Code Formatting` job. The aggregate `all-tests-passed` check is therefore red too.

One additional non-blocking limitation should be documented/tracked: a cached generic Boolean plan errors if `default_text_search_config` changes, while the freshly planned equivalent falls back to a valid sequential scan. The executor error preserves correctness, so I would not block this PR solely on that behavior, but prepared-statement users need to know to deallocate/reprepare after changing the configuration.

Coverage note: `boolean_concurrent_merge.sh` was added to the Makefile, but the main CI workflow manually enumerates shell tests and does not invoke it, so the current green PG17/18 jobs did not run that new concurrency test.

the-osiris:
Follow-up fixes are in [`4e6cd7e8`](https://github.com/the-osiris/pg_textsearch/commit/4e6cd7e8):

- **Mixed-negative warnings:** Fixed. We now use `tsquery_requires_match()`, so queries like `alpha | !missing` warn on legacy indexes.
- **Memtable locks/memory:** Fixed. Boolean scans now snapshot the chain under the lock, then stream one document at a time after releasing it.
- **Segment cancellation:** Fixed. Boolean exact-term iterators now use copied access, so they don’t retain buffer locks.
- **Formatting:** Fixed. `make format-check` passes.
- **Concurrency CI:** Added. The spill, merge, and VACUUM concurrency test now runs in CI.
- **Cached plans:** Documented. Cached Boolean plans must be prepared again after changing `default_text_search_config`.
- **Multi-segment tests:** Added coverage for multiple persisted segments plus a live memtable.
- **README documentation:** Added Boolean syntax, recheck behavior, scan modes, and text configuration notes.
- **Performance numbers:** I have the GIN comparison results ready and will add the short summary to the PR description.

tjgreen42:
The bounded memtable snapshot looks like a good correctness-first solution and should not block this PR. One follow-up optimization is worth tracking: Boolean execution currently bypasses the warm memtable cache and decodes every document vector in the captured chain for each query, whereas ranked BM25 queries can reuse cached postings.

I filed #499 to investigate a hybrid path: cache-backed posting evaluation for positive exact-term trees, with the bounded chain snapshot retained for full-universe, prefix/weight, recheck, and consistency-fallback cases. The issue explicitly preserves the snapshot design's bounded-memory, cancellation, and concurrent-reclamation guarantees.

tjgreen42 (approved):
LGTM, let's merge!  Will you do the honors?

PR #487, opened 2026-09-07
Bind the filtered top-k seed per index scan (#435)

### LINKED ISSUE #435: filtered_seed: key the top-K limit stash by scan identity (multi-scan-same-index)
Follow-up from PR #434 (selectivity-seeded top-K for filtered BM25 search).

## Problem

The seeded internal top-K is passed from planner (`tp_costestimate`) to executor (`tp_rescan`) through a single per-backend, per-`index_oid` slot (`tp_current_limit` in `src/index/limit.c`), which is consumed-and-cleared on first read.

Before #434 the stored value was always the raw `LIMIT k`, so multiple BM25 scans of the **same index** in one statement stored identical values — cross-contamination was invisible. #434 makes the stored value depend on each scan's own filter selectivity (`ceil(margin * k / s)`), so sibling scans of the same physical index with **different** filters now compute different seeds into that one shared slot. The last-planned seed wins; on execution the first scan consumes it and later siblings fall back to `tp_default_limit`.

Example shape (close to the faceted use case #434 targets):

```sql
SELECT id FROM docs WHERE facet = 'rare'   ORDER BY body <@> q LIMIT 10
UNION ALL
SELECT id FROM docs WHERE facet = 'common' ORDER BY body <@> q LIMIT 10;
```

## Impact

**Performance only — no correctness impact.** The executor Filter + Limit + the existing backoff re-drive still guarantee the exact correct top-k regardless of the slot's value. A mis-seeded arm just over-scores (bounded by `TP_MAX_QUERY_LIMIT`) or falls back into the backoff loop the seed was meant to avoid. Limited to the narrow shape of 2+ BM25 top-k scans over the same index in one statement with differing selectivities (UNION ALL arms, self-joins, a CTE referenced twice).

## Proposed fix

Key the limit hand-off by scan identity rather than `index_oid` alone (e.g. attach the seed to the plan/scan node at plan time, or otherwise disambiguate sibling scans), so each scan receives its own seed. This is a refactor of pre-existing `limit.c` infrastructure (the single-slot design deliberately replaced an older hash table "for simplicity") and was intentionally deferred out of #434.

Flagged by the automated review on #434.

mory91:
Fixes #435.

## The problem

`pg_textsearch.filtered_seed` (#434) seeds a BM25 scan's internal top-K
from the estimated selectivity of the filter above it, so a
`WHERE ... ORDER BY score LIMIT k` query can usually find its top-k in
one scoring pass instead of several executor backoff re-drives.

The seed was handed from planner to executor through a single
per-backend slot keyed only by `index_oid`, and the read consumed it:

```c
if (tp_current_limit.index_oid == index_oid)
{
    result = tp_current_limit.limit;

    /* Clear the limit after retrieval to prevent stale data */
    tp_current_limit.is_valid = false;
}
```

So in a statement with several BM25 scans of the same index — a faceted
`UNION ALL`, for instance — only the first scan to reach `tp_rescan` got
a limit at all. The rest fell back to `default_limit`. Whether that
happened to be deeper or shallower than the right depth was luck, which
is why seeding could make a multi-scan query *slower* than not seeding
at all (see the benchmark below).

## The approach

Seeds are now bound per scan at executor start rather than stashed at
plan time.

An `ExecutorStart_hook` walks the `PlanState` tree, carrying the nearest
enclosing `Limit` down through node types that preserve row identity and
order (an allow-list; anything else stops propagation). For each BM25
index scan it computes that scan's own seed from its own `k` and its own
filter selectivity, recovering the selectivity the planner already
estimated as `plan_rows / reltuples`.

The one wrinkle is where to put the seed so the access method can find
it. `index_beginscan` is called lazily from `IndexNext` on the first
tuple fetch, so at `ExecutorStart` there is no `IndexScanDesc` and no
`TpScanOpaque` to write to. What does exist is the scan's
`iss_OrderByKeys` array, so the binding is keyed by that array's address
plus the index OID, in a backend-local hash torn down by an
`es_query_cxt` reset callback. That address is stable across rescans,
which is what makes a correlated SubPlan re-seed correctly on every
evaluation.

Results are identical either way — the seed only sets scan depth, and
the executor's `Filter` plus backoff re-drive remain the correctness
safety net. Because depth is invisible in query results, a new
`bm25_debug_scoring_passes()` reports a session-local pass count, which
is what the tests assert on.

Design notes and the rejected alternatives are in
[`docs/issue_435_filtered_seed_scan_identity.md`](docs/issue_435_filtered_seed_scan_identity.md).

## Benchmark

`benchmarks/run_filtered_seed.sh` (new), 200k docs, PG 18, median of 7
runs. Comparing **this branch against `main` with the default
`filtered_seed = on`**, which is the change a user would actually feel:

| shape | selectivity | LIMIT | scans | main | branch | change |
|---|---|---|---|---|---|---|
| single | 0.001 | 10 | 1 | 8.95 ms | 8.76 ms | — |
| single | 0.001 | 100 | 1 | 26.21 ms | 25.91 ms | — |
| single | 0.01 | 10 | 1 | 2.28 ms | 2.21 ms | — |
| single | 0.1 | 10 | 1 | 1.64 ms | 1.61 ms | — |
| union2 | 0.001 | 10 | 2 | 18.7 ms | 18.0 ms | ~4% |
| union3 | 0.001 | 10 | 3 | 28.6 ms | 26.8 ms | ~6% |
| union3 | mixed | 10 | 3 | 81.0 ms | 58.5 ms | **~28%** |

Single-scan queries are untouched, as expected — one scan per statement
never collided, and that case was already handled by #434.

The gain concentrates in exactly one shape: several scans of one index
whose arms have *different* selectivities. When the arms share a
selectivity, main's single slot happens to hold a usable value for the
first arm and `default_limit` is not a bad guess for the others, so the
collision is close to benign and the win is only 4–6%.

The mixed-selectivity row is the one that argues for merging, for a
reason beyond its 28%: on `main` that query is **9% slower with
`filtered_seed = on` than with it off** (81.0 ms vs 74.2 ms) — the
feature is a net pessimization there. On this branch it is a 1.39x
speedup over off (58.5 ms vs 80.8 ms). Per-scan pass counts confirm the
mechanism: 23 passes → 5 for `union3 mixed`, 30 → 3 for `union3`
uniform, 10 → 1 for a single scan.

One caveat on reading the benchmark: the `seed off` column is **not**
comparable across the two builds, because "off" still binds the raw
user `k` and the two builds bind it to a different number of scans. Only
compare the `seed on` columns across versions.

## Testing

- 77/77 SQL regression tests pass on PG 18 (rebased onto `main` @
  f940210f, which landed background compaction).
- New `filtered_seed` cases cover one seed per arm in a multi-scan
  statement, correlated SubPlans re-seeding per evaluation, generic
  plans where the `LIMIT` is a `PARAM_EXTERN`, and the `LIMIT INT_MAX`
  saturation boundary (which must *not* seed — seeding it would drive a
  multi-GB allocation).
- Changed files pass CI's pinned clang-format 21.1.8.

Not addressed here: `so->limit` still sizes an unbounded allocation, so
a very large explicit `LIMIT` can fail on `palloc` independently of
seeding. That predates this change and capping it would silently
truncate results; written up as a follow-up in §13 of the design doc.

## Is this worth merging?

I want to flag the cost/benefit honestly rather than assume it.

**The cost** is roughly 570 lines of new C, a second executor hook, and
a mechanism that keys on the address of an executor-internal array with
its lifetime managed by a memory-context reset callback. It is well
tested and I believe it is correct, but it is more machinery than the
one-slot stash it replaces, and it is machinery a future reader has to
understand before touching the scan path.

**The benefit** is nothing for single-scan queries, 4–6% for multi-scan
queries with uniform arms, and ~28% plus the removal of a pessimization
for multi-scan queries with mixed arms.

There is a much cheaper fix that captures part of this: stop consuming
the slot on read, so every arm sees the last-planned seed instead of
falling back to `default_limit`. That is nearly a one-line change and
would get the uniform-arm cases (the 4–6%), but it leaves mixed arms
wrong, since one seed still has to serve scans that need different
depths — including the case where seeding is currently worse than
nothing.

So the question is whether the mixed-selectivity multi-scan shape is
common enough in real workloads to justify per-scan binding, or whether
the cheap fix (or narrowing `filtered_seed` to single-scan plans) is the
better trade. I lean toward merging, mainly because a feature that is
slower than its own `off` setting for a legitimate query shape seems
worth removing rather than documenting — but I do not have workload
evidence on how often that shape occurs, and I would rather defer to a
reviewer who does.


tjgreen42:
@mory91 This PR looks complex for the problem being addressed.  Can you simplify?

mory91:
I’ve implemented an alternative that carries the seed hint in a private copy of the `bm25query` value, using the existing planner hook and index-scan infrastructure. I’ve preserved it as a separate commit, [b26fed76](https://github.com/timescale/pg_textsearch/pull/487/commits/b26fed76), so the two approaches are easy to compare.

This removes the additional executor hook, per-scan registry, cleanup callbacks, and dedicated seed source files. In the seven benchmark cases tested, both approaches produced identical scoring-pass counts and broadly similar latency. The benchmark report is included in `docs/issue_435_benchmark_20260917.md`.

The tradeoff is narrower coverage: this version handles direct `Limit → IndexScan` plans when the query and limit are constants at planning time. Other cases use the existing backoff behavior. There are also two outstanding review findings around constant copying and prepared-plan detection that need fixing before this is ready.

My current understanding is that the difficult part isn’t calculating the seed—it’s reliably associating each scan with its own limit and filter estimate across prepared queries, rescans, and different plan shapes. Moving the hint into the query simplifies that bookkeeping, but introduces its own constraints.

I haven’t found an approach that both stays substantially simpler and preserves the original coverage. I’m open to suggestions, particularly if there’s an existing PostgreSQL or pg_textsearch mechanism that could carry this information more directly.

tjgreen42 (changes requested):
Review by **GitHub Copilot CLI**.

The planner-attached query-hint direction is substantially simpler than the original executor registry, and the conservative direct \`Limit -> IndexScan\` scope plus normal backoff preserves correctness. The multi-scan, correlated SubPlan, prepared-plan, and saturation coverage is strong. I think this is worth pursuing, but I found two issues to address before merge:

1. **Initialize the complete seed-hint representation** (\`src/types/query.c:1288-1297\`). \`TpQuerySeedHint hint\` has padding between \`magic\` and \`k\` on 64-bit builds, and the entire structure is copied into a varlena \`Const\`. PostgreSQL copies, compares, and serializes the complete raw datum, so the uninitialized padding makes the planned representation nondeterministic and copies stack bytes. Please zero-initialize it (for example, \`TpQuerySeedHint hint = {0};\`) and add a plan copy/serialization round-trip test.

2. **Avoid retaining abandoned query datums in cached-plan memory** (\`src/planner/hooks.c:1996-2006\`). \`copyObject(right)\` duplicates the original variable-length datum, then that copy's \`constvalue\` is overwritten by the hinted allocation; replacing the original right-hand node also leaves the original node/datum unreachable. Generic plans retain those allocations until invalidation or \`DEALLOCATE\`, which can be material for large query strings or multiple scan arms. Please construct the replacement \`Const\` once without the intermediate duplicated datum.

The branch also currently conflicts with \`main\`. In particular, \`query.c\` has changed substantially in #504 and #505, and the unrelated \`compaction_ownercheck_source.sh\` portability edit now conflicts with that script's deletion and should be dropped during the rebase. Please rebase and rerun CI against current \`main\`.

Finally, the PR description still documents the superseded \`ExecutorStart_hook\`/registry implementation rather than the current planner-attached query-value hints. Please update it so reviewers and future readers see the design that would actually merge.

PR #501, opened 2026-09-21
Make query optimization decition bound to query lifetime

marty1885:
Hello! This one is difficult to figure out. I was working on my search engine and found queries sometimes are slow.. randomly, for large, common words. So a wild hunt begins. Turns out

1. pg_textsearch recorded "this query has BM25" in one global boolean during parsing.
2. Parsing another statement could reset that boolean before PostgreSQL planned the prepared search. Which happens a lot for common terms
3. The planner skipped the rewrite that replace the `ORDER BY <@>` expression with the index’s cached score.
4. Evaluating the remaining expression re-tokenized returned documents—causing the long-word notices and slow searches.

So trying to run

```sql
PREPARE q(text) AS
    SELECT body FROM docs
    ORDER BY body <@> to_bm25query($1, 'docs_bm25_idx')
    LIMIT 10;

SELECT 1;
EXPLAIN (VERBOSE, COSTS OFF) EXECUTE q('hello');
```

Shows `body <@> ...` in the index scan's output buffer, and query is slower then expected.

This patch makes the decision per-query to avoid the issue.


tjgreen42:
@claude review

tjgreen42 (changes requested):
Copilot review — one blocker:

`collect_explicit_indexes_walker()` does not recurse into nested `Query` nodes: `query_tree_walker()` hands them to the callback, then `expression_tree_walker()` treats `T_Query` as a leaf. BM25 operators only in a CTE, FROM subquery, or `SubLink` leave `query_has_bm25_operators = false`, so score rewriting and explicit-index validation are skipped.

Reproduced on PG17: PR HEAD retains the raw `<@>` expression in a non-flattenable FROM subquery, while `origin/main` rewrites it to `bm25_get_current_score()`. Please recurse with `query_tree_walker()` for nested `Query` nodes and add nested-query EXPLAIN coverage.

— Copilot

tjgreen42:
Thanks for this report @marty1885!  Can you address the AI's feedback before I take a look at this?

PR #502, opened 2026-09-22
Add fair index writer admission

tjgreen42:
## Summary

PlanetScale's [TIN benchmark](https://planetscale.com/blog/introducing-tin) exposed writer starvation in the per-index lock. This PR establishes the locking foundations for nonblocking compaction:

- Waiting writers take priority over later readers, preventing a steady stream of searches from indefinitely delaying spill and maintenance work.
- Failed or interrupted writers unregister themselves and wake blocked readers, so cancellation cannot leave the admission gate stuck closed.
- Spill paths release the per-index lock before starting compaction, preserving a consistent maintenance-before-index lock order.
- Backend shutdown performs only best-effort spill work when relation or per-index access is busy, allowing process exit to complete instead of blocking behind concurrent index work.
- Explicit compaction fails with `lock_not_available` rather than waiting for maintenance admission or exclusive index access, avoiding lock cycles and keeping `bm25_compact_step()`'s `false` result unambiguous.

First of five; #504 is stacked on this PR.

## Validation

- PostgreSQL 17 and 18 regression suites
- PostgreSQL 18 concurrency suite
- Focused inline-compaction and shutdown-spill tests
- Source guards and formatting check


src/access/am.h
```diff
@@ -181,8 +181,9 @@ struct IndexBulkDeleteResult *tp_vacuumcleanup(
 char *tp_buildphasename(int64 phase);
 
 /*
- * Spill a memtable to an L0 segment.  Skips when
```
tjgreen42: Is this change needed?

